PR #58, opened 2026-04-10
Add versioned schema migrator (issue #43)

abl030:
## Summary

Closes #43.

Schema changes used to live inside `PipelineDB.init_schema()` as one giant idempotent SQL blob, applied as a side effect of any caller passing `run_migrations=True`. The recurring `soularr.service` path passed `run_migrations=False`, so deploys could ship code that depended on a new table without anything ever creating it on the live DB until an operator happened to run `pipeline-cli` (which still passed `run_migrations=True`). `search_log` in #41 hit exactly this failure mode.

This PR replaces that with a tiny **versioned migrator**:

- `migrations/NNN_name.sql` — applied in version order, exactly once per DB
- `lib/migrator.py` — ~150 lines: discover + apply, transactional per file, idempotent across runs, recorded in a `schema_migrations` tracking table
- `scripts/migrate_db.py` — thin CLI driven by the systemd unit
- `migrations/001_initial.sql` — frozen baseline containing everything `init_schema()` used to do (idempotent so it's safe against the live DB)

`PipelineDB` no longer runs DDL. The `run_migrations` kwarg, `init_schema()` method, and `SCHEMA_SQL` constant are gone. All call sites (pipeline_cli, repair, web/server, tests) drop the kwarg. `tests/conftest.py` applies migrations once at session start, so test setup helpers just `TRUNCATE`.

The deploy wiring lands in a **separate nixosconfig PR** that:
- Adds `pipeline-migrate` package + `soularr-db-migrate.service` (oneshot, `restartIfChanged = true`, `RemainAfterExit = true`)
- Wires `soularr.service` and `soularr-web.service` to `requires` it
- Pulls in this commit via `nix flake update soularr-src`

So every `nixos-rebuild switch` will run any pending migrations BEFORE the app services start. A failed migration blocks the app from coming up against an inconsistent schema.

## Why a custom migrator instead of Alembic / golang-migrate

- ~150 lines of Python, zero new deps (psycopg2 already used)
- Plain `.sql` files — no DSL, no SQLAlchemy models
- Fits the project's "stdlib + psycopg2" vibe
- Each migration is reviewable as a discrete file in the PR diff
- Versioned + tracked in a real table, with audit timestamps

## Adding a schema change in the new world

1. Drop a new file in `migrations/` named `NNN_describe_change.sql` (next number)
2. Plain SQL — each file runs in its own transaction, exactly once per DB. No `IF NOT EXISTS` guards needed.
3. `nix-shell --run "python3 -m unittest tests.test_migrator -v"`
4. Commit, push, deploy. The migrator picks it up automatically.

Documented in `.claude/rules/pipeline-db.md`, `CLAUDE.md`, `README.md`, the `/deploy` skill, and `.claude/rules/deploy.md`.

## Test plan

- [x] 1414/1414 tests pass
- [x] pyright clean on all touched files
- [x] New `tests/test_migrator.py` covers: file discovery (ordering, malformed names, duplicates), version recording, idempotency, partial-apply, rollback-on-failure, baseline-already-applied no-op
- [x] All existing test suites still green after `PipelineDB` no longer runs DDL (conftest applies migrations once)
- [ ] Deploy via the nixosconfig follow-up PR; verify `pipeline-cli query "SELECT * FROM schema_migrations"` shows version 1 on prod
- [ ] Verify `soularr-db-migrate.service` is `active (exited)` on doc2 after rebuild
- [ ] Trigger `soularr.service` manually, confirm it opens the DB cleanly


abl030:
Addressed two review issues on this branch:
- Fixed test DB setup so `tests/conftest.py` applies migrations for both externally supplied `TEST_DB_DSN` and the ephemeral Postgres path. Without this, CI or local runs with a pre-provided empty test DB could instantiate `PipelineDB` against an unmigrated schema.
- Tightened migration filename validation to require at least the documented three-digit numeric prefix (`NNN_name.sql`) while still allowing high-numbered test/future migrations such as `9001_*`.

Verification:
- `nix-shell --run "python3 -m unittest tests.test_migrator -v"`
- `nix-shell --run "python3 -m unittest tests.test_pipeline_db tests.test_album_source tests.test_force_import tests.test_pipeline_cli -v"`
- `nix-shell --run "pyright lib/migrator.py tests/conftest.py tests/test_migrator.py"`
- `nix-shell --run "bash scripts/run_tests.sh"` (1414 tests, 53 skipped)

Residual risk:
- Issue #43 is only fully satisfied once the companion deploy wiring in `abl030/nixosconfig#203` is merged, the `soularr-src` flake input is bumped to this PR, and doc2 is rebuilt/verified.

Issue #68, opened 2026-04-11
Web Decisions tab: render rank_gate_min_rank and rank_bitrate_metric as labeled badges


abl030:
Follow-up from #60 / #63 (codec-aware quality ranks).

`/api/pipeline/constants` now surfaces `rank_gate_min_rank` and `rank_bitrate_metric` (locked by contract test in `tests/test_web_server.py`), but the frontend Decisions tab at `web/index.html` doesn't render them. The decision tree visualization shows the FLAC/MP3 branching paths and stage rules, but operators can't see the configured rank policy at a glance.

## Scope
- `web/js/decisions.js` (or wherever the Decisions tab renders): add a badge row at the top of the tab showing:
  - **Gate min rank**: `EXCELLENT` (or whatever cfg says)
  - **Bitrate metric**: `avg` (or `min`)
  - **Within-rank tolerance**: `5 kbps` (or whatever cfg says)
- Pull from `state.constants.rank_gate_min_rank` etc — already in the constants payload, just unrendered
- Style: small inline pills, top-of-tab, like the existing threshold badges

## Acceptance
- Decisions tab shows live rank policy from the deployed `config.ini` without a page reload (re-fetched with the rest of the constants)
- No new API call needed

abl030:
Shipped in #76 (merged in ef787af). The Decisions tab now renders a three-pill rank policy badge row at the top — **Gate min rank** / **Bitrate metric** / **Within-rank tolerance** — pulled from the same `_runtime_rank_config()` snapshot the backend uses. Live on music.ablz.au as of deploy 2026-04-11.

Acceptance criteria from the issue:
- [x] Decisions tab shows live rank policy from the deployed `config.ini` without a page reload (re-fetched with the rest of the constants; `ef787af` added refetch-on-tab-reopen so retunes show up without full reload)
- [x] No new API call needed — uses the existing `/api/pipeline/constants` payload
- [x] Small inline pills at the top of the tab

Within-rank tolerance was added alongside the originally-scoped two fields because it's the third piece of the rank policy surfaced to the UI — all three come from one cfg snapshot now.

PR #76, opened 2026-04-11
Render rank policy as labeled badges on Decisions tab (#68)

abl030:
## Summary

Frontend follow-on to #75 — the API surface for `/api/pipeline/constants` is now fully cfg-coherent (thanks to the cfg threading), so this PR ships the UI payoff from issue #68: a labeled badge row at the top of the Decisions tab showing the live rank policy the deployed `config.ini` is running.

Three pills: **Gate min rank** / **Bitrate metric** / **Within-rank tolerance**. No new API round-trip — the constants payload already fetched by `loadDecisions()` now carries all three fields.

## What changed

**Backend** (`web/routes/pipeline.py`)
- `get_pipeline_constants` surfaces `rank_within_tolerance_kbps` alongside the existing `rank_gate_min_rank` and `rank_bitrate_metric` — all pulled from the same `_runtime_rank_config()` snapshot used for the threshold constants in #75.

**Frontend** (`web/js/decisions.js`, `web/index.html`)
- New exported pure function `renderPolicyBadges(constants)` returns an HTML string with three `.dp-policy-badge` pills. Called from `renderSimulatorForm` right after the "Decision Pipeline" title.
- Every interpolated value routes through `esc()`. The static `title` attribute has a maintainer-facing comment flagging that any future interpolation into it must also be escaped.
- Falsy-trap guarded: `!= null` not `||`, so a legitimately-configured `0 kbps` tolerance renders as `"0 kbps"` rather than `"?"`.
- Missing / null / undefined payload renders `"?"` placeholders so the function stays safe during boot and cache invalidation.
- Latent dead code cleaned up: `const c = state.dsConstants.constants` in `renderSimulatorForm` was unused before; now consumed by the badges call (scope.md "clean as you go").
- CSS: `.dp-policy` flex row with wrap for mobile, pill with label-on-top / value-on-bottom split, accent color `#6a9` matching `.dp-section-title`.

**Tests**
- `tests/test_web_server.py:test_pipeline_constants_contract` asserts `rank_within_tolerance_kbps` is present AND is an `int` — type pin so the frontend renders without conversion. Added RED-first, confirmed failing, then GREEN via the backend field.
- `tests/test_js_decisions.mjs` — new pure-function JS test file, **22 assertions**:
  - Happy path (default cfg)
  - Custom cfg (MEDIAN metric + `GOOD` gate + tolerance=12)
  - Zero-tolerance guard (falsy-trap regression)
  - Missing / null / undefined payloads
  - XSS case: `<script>alert(1)</script>` as gate rank must be escaped to `&lt;script&gt;`
- `scripts/run_tests.sh` runs the new JS test file alongside `test_js_util.mjs`.

**Doc**
- `CLAUDE.md` Decisions tab description updated to mention the badge row and cite #68.

## Pre-commit review gate

Ran an Opus review agent before committing (per `.claude/rules/code-quality.md`). Zero `[BLOCK]` findings. One `[NIT]` addressed (title-attribute escaping comment). The agent confirmed:
- Route already classified in `TestRouteContractAudit.CLASSIFIED_ROUTES`
- Pure-function testability: `decisions.js` imports `state.js` but state.js only touches DOM inside function bodies, so Node import works
- Contract test pins both presence and type
- Falsy trap correctly avoided
- XSS escaping in place with defense in depth

## Test plan
- [x] `nix-shell --run "bash scripts/run_tests.sh"` → 1522 Python tests OK (skipped=53), 37 + 22 JS tests pass
- [x] `nix-shell --run "pyright web/routes/pipeline.py tests/test_web_server.py"` → 0 errors
- [x] `node --check web/js/*.js` → clean on all files
- [ ] Deploy to doc2, reload Decisions tab, confirm the three badges render with the live `config.ini` values (gate=EXCELLENT, metric=avg, tolerance=5)
- [ ] Retune `config.ini` `within_rank_tolerance_kbps` to a non-default value (e.g. 10), deploy, confirm the badge updates on refresh — end-to-end proof that the UI tracks runtime cfg

## Acceptance (from issue #68)
- [x] Decisions tab shows live rank policy from the deployed `config.ini` without a page reload — re-fetched with the rest of the constants on tab open
- [x] No new API call needed
- [x] Rendered as small inline pills at the top of the tab

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -432,7 +432,7 @@ Uses `sox` bandpass filtering to detect transcodes. Measures RMS energy in 16 x 
 
 - **Recents tab** ("validation pipeline log"): Shows every download with full quality flow (slskd reported → actual on disk → spectral → existing). Badges: Upgraded, New import, Wrong match, Transcode, Quality mismatch. "On disk (before)" shows pre-import state.
 - **Library tab**: Quality label per album (MP3 V0, MP3 320, etc.). Upgrade button. Accept button (sets avg bitrate for lo-fi edge cases). Intent toggle: Default / Lossless (keeps lossless on disk for specific albums).
-- **Decisions tab**: Pipeline decision diagram generated from `get_decision_tree()` — shows FLAC/MP3 branching paths, all stages and rules with live thresholds from the code. Top of the tab renders a rank policy badge row (gate min rank / bitrate metric / within-rank tolerance) pulled from the same runtime cfg the backend uses, so operators see what `[Quality Ranks]` in the deployed `config.ini` is actually running (#68). Interactive simulator calls `full_pipeline_decision()` via `/api/pipeline/simulate` with presets for known scenarios.
+- **Decisions tab**: Pipeline decision diagram generated from `get_decision_tree()` — shows FLAC/MP3 branching paths, all stages and rules with live thresholds from the code. A rank policy badge row at the top of the tab (gate min rank / bitrate metric / within-rank tolerance, issue #68) mirrors the same runtime cfg the backend uses, so operators see at a glance what `[Quality Ranks]` in the deployed `config.ini` is actually running. Interactive simulator calls `full_pipeline_decision()` via `/api/pipeline/simulate` with presets for known scenarios.
 - **Ban source**: Denylists user + removes from beets + requeues.
 
 ### Edge Cases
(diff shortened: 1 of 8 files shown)


abl030:
Addressed the stale Decisions-tab config issue on this branch:
- removed the one-shot client cache so opening the Decisions tab re-fetches `/api/pipeline/constants` each time
- added a JS regression test that opens the tab twice with different payloads and asserts the second render shows the fresh rank policy

Verification:
- `node tests/test_js_decisions.mjs`
- `node --check web/js/*.js`

Residual risk:
- none beyond the existing behavior that rapid repeated tab clicks can issue repeated GETs, which is acceptable here because `/api/pipeline/constants` is not server-side cached and is cheap to compute.

PR #100, opened 2026-04-18
refactor(quality): msgspec.Struct at the harness wire boundary (#99)

### LINKED ISSUE #99: Strict-typed JSON validation at the harness wire boundary (msgspec)
## Motivation

PR #98 + the follow-up commit `5c01ba7` fixed an int-vs-str MBID bug that caused **every Discogs validation to log `mbid_not_found` and reject**. Beets' Discogs plugin returns `album_id` as a JSON number; our typed dataclass said `str`; consumers compared with `==` against the str DB-stored `mb_release_id`. The fix coerced at the harness boundary plus belt-and-braces in three consumer sites.

That whole class of bug — typed dataclass field says `X`, runtime JSON delivers `Y`, equality silently fails — is invisible to pyright because pyright can't see inside `dict.get()`. **A strict-typed deserializer at the wire boundary would have caught it at decode time, in one place, with no defensive coercion needed downstream.**

## Proposal

Replace the hand-rolled `from_dict` / `from_json` / `to_json` methods on the harness wire-boundary dataclasses with [`msgspec`](https://jcristharif.com/msgspec/). msgspec is essentially "dataclasses + free typed JSON validation":

- Declare `class Foo(msgspec.Struct): album_id: str`
- Decode with `msgspec.json.decode(blob, type=Foo)` — wrong type → `ValidationError` at the boundary
- Encode with `msgspec.json.encode(obj)`
- Near-zero overhead, no extra dataclass-unpacking step
- Already in nixpkgs (`python3Packages.msgspec`, version 0.20.0 on the current channel — confirmed via `nix-shell -p python3Packages.msgspec`)

### In scope (the actual wire boundary)

Four dataclasses in `lib/quality.py`:
- `HarnessItem` (line 228)
- `HarnessTrackInfo` (line 244)
- `TrackMapping` (line 261)
- `CandidateSummary` (line 276)

Plus the consumer in `lib/beets.py::beets_validate` (replaces the hand-rolled `_candidate_from_harness` + raw `json.loads` of harness stdout lines).

### Out of scope (consumer-side dataclasses, not wire-boundary)

Leave these alone unless someone wants to do them in a follow-up:
- `ValidationResult` (top-level dump that wraps `CandidateSummary` lists)
- `ImportResult`, `ConversionInfo`, `SpectralDetail`, `PostflightInfo` (the import-side dump)
- `ActiveDownloadState` / `ActiveDownloadFileState` (DB persistence, not harness wire)
- `DownloadInfo`, `AudioQualityMeasurement`, `QualityRankConfig`, etc.

These all serialize to JSONB columns or stdout sentinels, but their inputs are constructed in our own Python code — they don't deserialize from an external schema we don't control. The bug surface is the **harness boundary**, not the audit log.

## Migration concern (good news)

The `download_log.validation_result` JSONB column has 2656 live rows. **Audit on doc2:**

```
candidates_with_int_id: 0
tracks_with_int_id:     0
```

Zero rows have integer IDs in the candidate dump. Reason: every Discogs validation was rejected pre-fix before the candidate was logged, and every successful validation pre-fix was MusicBrainz (UUID strings). **No data migration needed.** Post-fix, the harness coerces to str before logging, so future rows are also clean.

If we extend scope to `ValidationResult`, decoding old log rows might trip on missing optional fields — `msgspec.Struct` allows defaults same as dataclasses, so a careful field-by-field audit would address it.

## Ballpark effort

- **Half-day** for the 4 wire-boundary dataclasses + `lib/beets.py` decoder + tests + Nix dep
- Nix changes: add `ps.msgspec` to `shell.nix:25-30` and to `nixosconfig/modules/nixos/services/soularr.nix:88-94` (one line each)
- Test churn: `tests/test_validation_result.py` (~408 lines, mostly unchanged — `Foo(**dict)` patterns still work because `msgspec.Struct` accepts kwargs)
- Risk: low — all changes are at one boundary, with full unit-test coverage already in place

## Acceptance criteria

- `lib/beets.py::beets_validate` decodes harness `choose_match` lines via `msgspec.json.decode(line, type=ChooseMatchMessage)` (or equivalent)
- `CandidateSummary.from_dict` / `_coerce_id` / `_coerce_track_ids` deleted — msgspec handles the type contract
- A regression test that feeds a harness JSON line with `album_id: 2085134` (int, no quotes) and asserts `msgspec.ValidationError` (or successful coerce, depending on which strategy) — proves the boundary is strict
- All 1660+ existing tests still pass
- `pyright` clean
- Optional: drop the `_id_str` helper in `harness/beets_harness.py` — the harness still emits int (we control the Python-side serialization there too, and beets' AlbumInfo carries native types). Decide whether to keep harness-side coercion as redundant defense or trust msgspec on the consumer side. **Recommend keeping** the harness-side `_id_str` for a single reason: it normalizes the wire format so when JSONB rows are queried via `pipeline-cli query` or read by the web UI, they always show str, regardless of whether msgspec is in the loop.

## Reference

- PR #98: `fix(beets): coerce Discogs int IDs to str at harness boundary`
- Commit `5c01ba7`: `fix(import_one): coerce Discogs int album_id at the import stage` (the same bug at a second consumer site — strict deserialization would have made this single-fix-only)
- msgspec docs: https://jcristharif.com/msgspec/

abl030:
Closes #99.

## Summary
- Replaces hand-rolled `from_dict` + defensive int→str coercion on the four harness wire-boundary dataclasses (`HarnessItem`, `HarnessTrackInfo`, `TrackMapping`, `CandidateSummary`) with `msgspec.Struct`. Adds `ChooseMatchMessage` so the whole `choose_match` line decodes in one strict-typed step at `lib/beets.py::beets_validate`.
- Turns the class of bug PR #98 fixed (typed dataclass says `str`, JSON carries `int`, equality silently fails → every Discogs validation logs `mbid_not_found`) from invisible-to-pyright drift into a loud `msgspec.ValidationError` at one well-defined boundary. The harness-side `_id_str` still normalises IDs to str before emitting so the strict boundary never trips in the happy path; it exists to catch regressions.
- Adds a new code-quality rule (`.claude/rules/code-quality.md` § "Wire-boundary types") documenting the policy for future wire-level deserialization: `msgspec.Struct` at boundaries, `@dataclass` for in-process types, normalise early (at the source), never coerce on the consumer side.

## Shape of the change
- **Deleted**: `_coerce_id`, `_coerce_track_ids`, `_candidate_from_harness`, `CandidateSummary.from_dict`, `TrackMapping.from_dict`.
- **`ValidationResult`** stays `@dataclass` (out of scope per the issue) but its `to_json` / `from_dict` / `from_json` now delegate to msgspec so the Struct children round-trip correctly.
- **`CandidateSummary`** uses `rename={"mbid": "album_id"}` so the JSON key matches beets' own naming while Python callers keep `.mbid` (no consumer-side churn).
- **Regression guards** in `test_validation_result.py` and `test_beets_validation.py` assert `msgspec.ValidationError` on int/null ID inputs — the detector that makes the boundary worth having. The live-bug reproduction test is now flipped: feeding an int `album_id` asserts `result.error` is set and names the offending field, rather than silently coercing to match.

## Out of scope (per issue #99)
- `ValidationResult`, `ImportResult`, `ConversionInfo`, `SpectralDetail`, `PostflightInfo`, `ActiveDownloadState`, `DownloadInfo`, etc. all stay `@dataclass` — their inputs are constructed entirely in our own Python code, so strict runtime validation buys nothing.

## Nix
- `ps.msgspec` added to `shell.nix` (already on nixpkgs 0.20.0).
- `nixosconfig/modules/nixos/services/soularr.nix` needs the same one-line addition — will land as a separate commit on doc1 before merging/deploying so the module is updated in lockstep.

## Test plan
- [x] `nix-shell --run "bash scripts/run_tests.sh"` → 1672 tests, all pass (53 skipped as usual)
- [x] `pyright lib/quality.py lib/beets.py tests/test_validation_result.py tests/test_beets_validation.py` → 0 errors
- [x] New test `test_int_album_id_trips_msgspec_boundary` in `test_beets_validation.py` asserts the strict boundary catches an int `album_id` and surfaces it as `result.error` with the offending field name
- [x] New tests in `TestCandidateSummaryIdContract` assert `msgspec.ValidationError` on int/null ID inputs at every ID-typed field (mbid/releasegroup_id/track_id/release_track_id/nested HarnessItem.path)
- [x] New `TestStructIsMsgspec` pins the four types as `msgspec.Struct` subclasses
- [ ] Codex review (fix-bug skill) — scheduled before merge
- [ ] Deploy verification on doc2 after nixosconfig update

### DIFF
--- a/.claude/rules/code-quality.md
+++ b/.claude/rules/code-quality.md
@@ -8,6 +8,17 @@
 - Inner data structures must also be typed — no `list[dict]` when a dataclass exists
 - Verify with: `pyright <files>` on every touched file before committing
 
+## Wire-boundary types — use `msgspec.Struct`, not `@dataclass`
+Any type whose **input comes from outside our Python code** (harness stdout, an HTTP response, a JSONB blob coming back from the DB, a subprocess's stdout) must be a `msgspec.Struct` so the deserializer validates the typed contract at the boundary. Pyright does not see inside `dict.get()` — only runtime validation catches int-vs-str drift, mis-typed fields, or missing required data. This is the lesson of issue #99 / PR #98 (every Discogs validation silently logged `mbid_not_found` because a dataclass said `str` but the wire carried `int`).
+
+- **Use `msgspec.Struct`** for: harness/subprocess JSON messages, external API responses, DB JSONB rows we read back and type-check, any other "decode this blob into a typed object" site.
+- **Keep `@dataclass`** for: types we construct entirely from our own typed Python code (`ImportResult`, `QualityRankConfig`, `SoularrConfig`, `ActiveDownloadState` when we're the only writer). Their inputs are already typed — the strict boundary buys nothing.
+- **Decode at exactly one site.** The wire boundary is the one place the untyped blob becomes a typed object. After that, every downstream consumer works with the Struct directly — no defensive coercion, no `dict.get()`, no re-validation. If you find yourself writing a `_coerce_x` helper on the consumer side, the boundary is in the wrong place.
+- **Strict ≠ coerce.** Declare fields as the type you want (`str`, not `str | int`). `msgspec.ValidationError` at the boundary is the detector. Do not use `strict=False` to silently coerce away real type drift.
+- **Normalise early if the external source is untidy.** Harness-side `_id_str` in `harness/beets_harness.py` coerces int IDs to str *before* emitting — so the wire is always clean, and the downstream Struct validation never trips in the happy path. Keep normalisation at the source, not at the consumer.
+- Reference implementation: `HarnessItem`, `HarnessTrackInfo`, `TrackMapping`, `CandidateSummary`, `ChooseMatchMessage` in `lib/quality.py`; decoded at `lib/beets.py::beets_validate` via `msgspec.convert(msg, type=ChooseMatchMessage)`.
+- Tests owe: at least one RED test that feeds the wrong type at the boundary and asserts `msgspec.ValidationError`. This is the regression guard that makes the boundary worth having.
+
 ## Testing — Red/Green TDD
 - Write tests FIRST (RED), then implement (GREEN)
 - Every new function, dataclass, and decision branch needs test coverage
--- a/.claude/rules/harness.md
+++ b/.claude/rules/harness.md
@@ -9,6 +9,7 @@ paths:
 
 - The harness runs in the beets Python environment (Nix Home Manager on doc1), NOT in the dev shell
 - `_serialize_album_candidate()` must capture EVERY field from AlbumMatch — never discard data
-- All harness output types must be typed dataclasses: HarnessItem, HarnessTrackInfo, TrackMapping
+- **Harness-emitted types are `msgspec.Struct`, not `@dataclass`** — `HarnessItem`, `HarnessTrackInfo`, `TrackMapping`, `CandidateSummary`, `ChooseMatchMessage`. Decoded at the wire boundary in `lib/beets.py::beets_validate` via `msgspec.convert(msg, type=ChooseMatchMessage)`. The strict-typed decoder is what catches int-vs-str drift (PR #98 bug, issue #99). See the `Wire-boundary types` section in `.claude/rules/code-quality.md` for the full policy.
+- **Never normalise on the consumer side.** If beets emits an int where our Struct says str, fix `_id_str` in `harness/beets_harness.py`. Defensive coercion downstream defeats the boundary.
 - import_one.py emits ImportResult as a `__IMPORT_RESULT__` sentinel JSON line on stdout, human logging on stderr
 - Use `beets-docs` skill (`.claude/commands/beets-docs.md`) to look up beets internals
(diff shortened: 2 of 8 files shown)


abl030:
## Codex review round 1

Codex found one **P1** issue:

> Wire msgspec into the service and harness runtimes — shell.nix:28. When this branch is run via the actual import/validation paths instead of `nix-shell`, `lib/quality.py` and `lib/beets.py` now hard-import `msgspec`, but this diff only adds the package to the dev shell.

### Resolution

The infrastructure side of this change lives in the `nixosconfig` repo, not in this PR. I landed the companion change on doc1:

- `nixosconfig` commit [`c1cf30f2`](https://github.com/abl030/nixosconfig/commit/c1cf30f2): adds `ps.msgspec` to the soularr `pythonEnv` used by `soularr.service`, `soularr-web.service`, and `pipeline-cli`.

Verified scope:
- `harness/beets_harness.py` doesn't import `lib/quality.py`, so the beets (Nix Home Manager) env on doc1 doesn't need msgspec. Confirmed by `grep -n '^import\|^from' harness/beets_harness.py` — it only imports stdlib + beets.
- `harness/import_one.py` DOES import `lib.quality`, but `lib/import_dispatch.py` launches it via `sys.executable` (line 488) — that's the soularr service env, already updated in `c1cf30f2`.
- `soularr.py`, `soularr-web`, `pipeline-cli` all run under the same `pythonEnv` — same fix covers them all.

The deploy sequence is: merge this PR → `nix flake update soularr-src` on doc1 → `nixos-rebuild switch` on doc2. Both the msgspec env addition and the soularr code that uses it land atomically.

No code changes needed on this branch. Proceeding to merge.

Issue #84, opened 2026-04-14
Imported album dirs created as 0755 root:root despite UMask=0000 — blocks other users' beets ops


abl030:
## Summary

Albums imported via the soularr pipeline (auto-import path: `process_completed_album` → `import_one.py` → beets harness) end up as `drwxr-xr-x root:root` on disk, instead of the `drwxrwxrwx root:root` expected from the service's `UMask = "0000"`. The files inside the album dir are correct (`-rw-rw-rw- root:root`). Only the *dirs created by beets during import* are affected.

This blocks any non-root user from running subsequent beets operations (e.g. `beet fetchart`, `beet embedart`, re-tag, in-place edits) because beets needs to create sibling temp files inside the album dir.

### Dirs created by soularr's own code are correct (0777)

```
$ ls -ld /mnt/virtio/Music/Incoming/*/*
drwxrwxrwx root root ... /Incoming/Lanu/Her 12 Faces
drwxrwxrwx root root ... /Incoming/M83/Dead Cities, Red Seas & Lost Ghosts
drwxrwxrwx root root ... /Incoming/Parts & Labor/A Great Divide 7
...
```

These come from `lib/util.py::stage_to_ai()` → `os.makedirs(dest, exist_ok=True)`. ✓

### Dirs created by beets during import are wrong (0755)

```
$ find /mnt/virtio/Music/Beets -maxdepth 3 -type d \( -user root -o -group root \) | wc -l
410

$ stat '/mnt/virtio/Music/Beets/The Victims/2005 - All Loud on the Western Front'
Access: (0755/drwxr-xr-x)  Uid: (0/root)   Gid: (0/root)

$ ls -l '/mnt/virtio/Music/Beets/The Victims/2005 - All Loud on the Western Front/'
-rw-rw-rw- 1 root root  5.0M ...  05 I Understand.mp3
-rw-rw-rw- 1 root root  5.1M ...  06 Open Your Eyes.mp3
-rw------- 1 root root   13K ...  cover.jpg     # ← 0600, from Python tempfile
```

### Impact (today)

```
$ beet fetchart -q
...
PermissionError: [Errno 13] Permission denied:
  '/mnt/virtio/Music/Beets/The Victims/2005 - All Loud on the Western Front/.cover.jpg.2f1sajv_.beets'
```

[...]

PR #136, opened 2026-04-20
refactor: BeetsAlbumOp + release lock + sibling imported_path propagation (#133, #132)

### LINKED ISSUE #132: Post-PR #131 follow-ups: cross-process same-release lock + sibling imported_path propagation
Two correctness gaps surfaced by Codex during PR #131 review (round 6) that require plumbing larger than the blast-radius data-loss fix. Both are tagged as KNOWN LIMITATIONS in comments at the relevant callsites in \`harness/import_one.py\`.

---

## 1. Cross-process race in post-import stale cleanup

**Where**: \`harness/import_one.py::main\`, the post-import cleanup loop around line 1547 (\`new_album_id = max(post_import_ids)\`).

**The scenario**: we derive \"the album we just imported\" as the highest \`albums.id\` among same-MBID rows (beets auto-increments, so freshest = max). If a SECOND soularr process commits its own same-MBID row between our \`run_import\` returning and our re-enumerate query, \`max()\` picks the other process's row as \"new\" — our cleanup then deletes the one THIS process actually imported and logs success against the wrong files.

**How the race happens in practice**: 
- \`soularr.service\` is single-instance via \`/var/lib/soularr/soularr.lock\`, so two timer-driven cycles can't overlap.
- \`soularr-web\` (force-import / manual-import routes) does NOT hold that lock — it spawns \`import_one.py\` subprocesses independently.
- Therefore: user clicks force-import on album X via the web UI while soularr.service is mid-cycle importing album X → race.

**Proper fix**: pipeline-level same-release advisory lock shared by both entry points. Sketch:
- Acquire a lock keyed on \`mbid\` (PostgreSQL advisory lock, or a row in a \`release_locks\` table with \`ON CONFLICT DO NOTHING\`).
- \`dispatch_import_core\` acquires before spawning \`import_one.py\`, releases after it exits.
- Both \`soularr.py\` (main cycle) and \`web/routes/pipeline.py\` (force/manual) go through the same dispatch path, so one lock covers both.

**Impact scope**: narrow — requires concurrent imports of the same album from two different soularr entry points. In the race window, the user loses the new album's files and retains a stale row. Ban-source cleanup recovers. Not data loss, but silent wrong-state.

---

## 2. Sibling \`imported_path\` not propagated to pipeline DB

**Where**: \`harness/import_one.py::_canonicalize_siblings\`, after \`_run_album_move_by_id\` succeeds for a sibling.

**The scenario**: when beets' \`%aunique\` suffixes the new album, sibling editions get re-moved to canonical disambiguated paths too (e.g. \`/2006 - Palo Santo/\` → \`/2006 - Palo Santo [2006]/\`). \`_canonicalize_siblings\` handles the disk move and calls \`fix_library_modes\` for perms, but it does NOT update the sibling's \`album_requests.imported_path\` row in the pipeline DB.

**Consequence**: if the sibling was tracked as a prior pipeline request, its UI path is stale until some other event updates it. The user clicks the sibling in the UI → points at a nonexistent \`/2006 - Palo Santo/\` directory until next upgrade or re-tag.

**Proper fix**: cross-service plumbing so the harness (or the parent dispatch) can update \`album_requests.imported_path\` for any request whose beets album id appears in \`sibling_album_ids\` post-move. Currently \`import_one.py\` only writes to the pipeline DB for its own \`--request-id\`. Options:
- Extend \`import_one.py\` to look up pipeline requests by beets \`album_id\` and update their paths (widens its DB responsibilities).
- Return the list of moved sibling album ids + new paths from \`import_one.py\` as JSON, have \`dispatch_import_core\` do the pipeline DB updates (cleaner separation).

**Impact scope**: cosmetic for most users (UI path wrong until next event). No data loss. Only matters when the sibling edition was itself a tracked pipeline request — fresh imports of orphan siblings don't hit it.

---

## Not tackling

A third theme came up during review that's worth noting but NOT a correctness bug: beets' \`find_duplicates()\` returns cross-MBID siblings despite \`duplicate_keys: [albumartist, album, mb_albumid]\` being configured as AND. We confirmed this empirically in the live 2026-04-20 logs but didn't root-cause it. The blast-radius fix in #131 makes the upstream contract violation non-catastrophic — any false-positive duplicate now gets \`\"keep\"\`, not \`\"remove\"\`. Worth investigating at some point but not blocking.

### LINKED ISSUE #133: refactor: extract BeetsAlbumOp abstraction for beets subprocess operations
## Motivation

PR #131 (Palo Santo data-loss fix) shipped after six rounds of Codex review. Each round surfaced a distinct correctness leaf, but in aggregate they trace a single missing abstraction: **every beets subprocess operation that moves or removes files has the same set of invariants, and every callsite implements some subset and forgets the rest.**

The invariants (lifted from the six review rounds, distilled):

1. **Narrow scope.** Album-mode (`-a`), primary-key-scoped selectors where possible (`id:<N>` over `mb_albumid:<uuid>`).
2. **Source-agnostic.** Works for both MusicBrainz (`mb_albumid`) and Discogs (`discogs_albumid`) rows. The beets numeric `albums.id` is the one identifier always populated.
3. **Correctly ordered.** Destructive steps only fire after all reversible steps succeed. Pre-flight captures are non-destructive; destructive work happens only after the replacement is confirmed on disk.
4. **Race-safe.** Same-release concurrent imports (soularr.service vs. soularr-web force-import) must not delete each other's rows. Requires a pipeline-level same-release advisory lock.
5. **Permission-safe.** `fix_library_modes` runs after any `beet move` that may have created a fresh disambiguated directory (issue #84).
6. **Pipeline-DB-consistent.** When a sibling's path changes on disk, its `album_requests.imported_path` must update too, or the UI lies.

Today these live as ad-hoc checks at five+ callsites (pre-flight, post-import cleanup, `_apply_disambiguation`, `_canonicalize_siblings`, `remove_and_reset_release`). Each callsite has drifted: some miss `-a`, some miss perm repair, some miss DB propagation, some miss the race.

## Proposal: `BeetsAlbumOp`

A typed wrapper around `beet remove`/`beet move` that enforces all six invariants at the type level. Sketch:

\`\`\`python
@dataclass(frozen=True)
class BeetsAlbumHandle:
    album_id: int  # always populated; source-agnostic
    mbid: str  # may be \"\" for Discogs; informational

class BeetsAlbumOp:
    def __init__(self, beets_db, pipeline_db, lock_manager): ...

    def remove(self, handle: BeetsAlbumHandle, *, delete_files: bool) -> OpResult:
        # -a id:<N> selector; acquires same-release lock; updates
        # pipeline DB on success.

    def move(self, handle: BeetsAlbumHandle) -> OpResult:
        # -a id:<N>; runs fix_library_modes on new path; updates
        # pipeline DB if imported_path changed.
\`\`\`

Every current callsite routes through this. The callsite code becomes \"get a handle, call op\"; the invariant is in the op, not the callsite.

## In scope

- Extract `BeetsAlbumOp`, `BeetsAlbumHandle`, `OpResult` types.
- Migrate every callsite in `harness/import_one.py` and `lib/release_cleanup.py`.
- Add a same-release advisory lock in `lib/pipeline_db.py` (PostgreSQL `pg_advisory_lock` or a `release_locks` table), acquired by `BeetsAlbumOp`, held across the op.
- Propagate sibling `imported_path` to `album_requests` for every moved album id that maps to a tracked pipeline request (the harness currently only writes for its own `--request-id`).
- Retire `_capture_stale_beets_id`'s conceptual successor (the inline `stale_ids` enumerate), `_apply_disambiguation`'s hardcoded `mb_albumid`-based selector, and `_canonicalize_siblings`' scattered perm/DB logic — all replaced by `BeetsAlbumOp.move` / `.remove`.

## Out of scope

- The existing data-loss fix in #131 stays as-is. The refactor lands on top of a known-safe baseline so we can bisect if something breaks.
- Upstream beets' `find_duplicates()` returning cross-MBID siblings when `duplicate_keys` is AND — #131's blast-radius fix makes this non-catastrophic, so it's still parked as a separate investigation.

## Acceptance

- Every `beet remove`/`beet move` subprocess in the repo goes through `BeetsAlbumOp`.
- No callsite constructs its own `beet remove -a -d id:<N>` argv.
- A contract test asserts that grepping the codebase for `\"beet\", \"remove\"` or `\"beet\", \"move\"` finds matches only in `BeetsAlbumOp` and tests.
- Closes #132.
- Same-release lock covered by an orchestration test simulating concurrent imports of the same MBID.
- Sibling `imported_path` propagation covered by an integration slice.

## Context

- PR #131 commit history is the audit trail — each `review(round N)` commit is a leaf of this abstraction.
- `.claude/commands/fix-bug.md` has a \"stop and refactor\" rule after round 2 that I should have triggered and didn't. This issue is the deferred payment on that debt.

abl030:
## Summary

Closes #133 (the BeetsAlbumOp extraction — the refactoring-debt payment from PR #131's six rounds of Codex review) and #132 (both follow-ups: cross-process release lock + sibling imported_path propagation).

Three feature commits, each followed by an adversarial-review-fix commit, plus a final review pass. Rebase-merge keeps the audit trail on \`main\`.

### Commit 1 — \`BeetsAlbumOp\` abstraction

- New \`lib/beets_album_op.py\` with \`BeetsAlbumHandle\`, \`BeetsOpFailure\`, \`BeetsOpResult\`, and three entry points: \`remove_album(handle)\`, \`move_album(handle, beets_db)\`, \`remove_by_selector(selector)\`.
- All five+ callsites migrated: \`_apply_disambiguation\`, \`_canonicalize_siblings\`, \`_remove_stale_by_id_logged\`, \`remove_album_by_beets_id\`, \`remove_album_by_selectors\`.
- \`fix_library_modes\` (issue #84) moved INTO \`move_album\` — every caller gets perm repair, no exceptions.
- Contract test \`TestBeetOpArgvIsCentralised\` greps the repo for three argv shapes (\`"beet"\`, \`beet_bin()\`, \`BEET_BIN\`) and fails if any file outside the allowlist constructs raw argv.
- Unified \`DisambiguationFailure\` + \`SelectorFailure\` into \`BeetsOpFailure\` (kept name aliases). Back-compat preserved via default empty \`selector\` field — old JSONB rows deserialize cleanly; explicit test pins this.

### Commit 2 — Cross-process same-release advisory lock (#132 P1)

- New \`ADVISORY_LOCK_NAMESPACE_RELEASE\` keyed on \`zlib.crc32(mb_release_id.strip())\` (stable across processes, unlike salted \`hash()\`). Non-blocking acquire in \`dispatch_import_core\`.
- Closes the \`max(post_import_ids)\` cross-process race documented as KNOWN LIMITATION in harness/import_one.py (now also the comment is updated).
- **C1 regression fix** (caught by adversarial review): auto-path contention previously left \`status='downloading'\` → \`_run_completed_processing\` silently flipped the request to \`imported\` without importing. Now resets to \`wanted\` + cleans staging. Force/manual path preserves status.
- Orchestration slices for contention + happy path + empty-mbid fallback.

### Commit 3 — Sibling \`imported_path\` propagation (#132 P2)

- \`_canonicalize_siblings\` now returns \`list[MovedSibling]\`; records flow through \`PostflightInfo.moved_siblings\` and are propagated by new \`_propagate_moved_siblings\` in \`dispatch_import_core\`.
- New \`PipelineDB.update_imported_path_by_release_id(mb_albumid, discogs_albumid, new_path)\` — covers both layout combos.
- Strict wire-boundary validation: \`_postflight_from_dict\` decodes via \`msgspec.convert\` (issue #99 lesson). Regression test feeds \`album_id="10314"\` as string and asserts \`msgspec.ValidationError\`.
- Observability WARNs for the silent-failure modes (rowcount>1, both-ids-empty). Defensive try/except around \`get_release_ids_by_album_id\` so a beets lock doesn't kill the whole canonicalization loop.

## Commits

\`\`\`
1c73ab3 review(final): drop dead fields + fix doc lie + add fake self-tests
0110b9c review(commit 3): strict-type boundary + observability + defensive lookups
caa0d51 feat: propagate sibling imported_path to pipeline DB (#132 P2, #133)
43e83e8 review(commit 2): fix auto-path status-flip regression + 3 polish items
35e85ac feat: cross-process same-release advisory lock (#132 P1, #133)
b3d683c review(commit 1): widen argv grep, add pre-#133 JSON test, polish
81d3953 refactor: extract BeetsAlbumOp for beets subprocess operations (#133)
\`\`\`

## Acceptance (per #133)

- [x] Every \`beet remove\` / \`beet move\` subprocess routes through BeetsAlbumOp.
- [x] No callsite constructs its own argv — contract test enforces this.
- [x] Same-release lock covered by an orchestration test (\`TestReleaseLockContention\`).
- [x] Sibling \`imported_path\` propagation covered by an integration slice (\`TestSiblingImportedPathPropagation\`).
- [x] Closes #132.

## Test plan

- [x] \`nix-shell --run "bash scripts/run_tests.sh"\` — **1965 passing**, 53 skipped.
- [x] \`pyright\` — 0 errors on every touched file.
- [ ] Deploy to doc2 via \`/deploy\` flow. No migrations needed (\`discogs_release_id\` index already ships in \`002_discogs_index.sql\`).
- [ ] Post-deploy: verify a kept-duplicate import (search a multi-edition album) triggers the sibling propagation path and logs show \`[CANONICALIZE] … emitted for pipeline DB propagation\` + \`propagated imported_path →\` on the dispatch side.

### DIFF
--- a/harness/import_one.py
+++ b/harness/import_one.py
@@ -1446,18 +1446,16 @@ def main():
     # we no longer depend on the pre-flight capture matching what
     # ``resolve_duplicate`` sees. The intra-process race is closed.
     #
-    # KNOWN LIMITATION (Codex PR #131 round 6 P1): this remains
-    # vulnerable to CROSS-process races. If a second cratedigger process
-    # (e.g. cratedigger-web force-import racing an auto-import cycle)
-    # commits a same-MBID row between our ``run_import`` returning
-    # and this re-enumerate query, ``max(post_import_ids)`` would
-    # pick that other process's row as "new" and our cleanup would
-    # delete the one this process actually imported. Cratedigger.service
-    # is single-instance (``/var/lib/cratedigger/cratedigger.lock``) but the
-    # web force-import path does not hold that lock, so the race
-    # window exists in practice. Follow-up: a pipeline-level same-
-    # release advisory lock shared by both entry points, tracked
-    # separately from this data-loss fix.
+    # Cross-process race closed by issue #133 / #132 P1:
+    # ``dispatch_import_core`` takes ``ADVISORY_LOCK_NAMESPACE_RELEASE``
+    # keyed on a stable hash of ``mb_release_id`` for the duration of
+    # this subprocess. Previously ``max(post_import_ids)`` was
+    # vulnerable to a racing process inserting its own same-MBID row
+    # between ``run_import`` returning and this re-enumerate query;
+    # the release-level advisory lock serialises every entry point
+    # (auto cycle, web force-import, CLI manual-import) that spawns
+    # import_one.py so this window no longer exists. The non-blocking
+    # lock means a second caller returns early rather than queueing.
     #
     # Each removal uses ``beet remove -a -d id:<N>`` — primary-key
     # scoped, cannot reach cross-MBID siblings. Any removal failure
--- a/harness/import_one.py
+++ b/harness/import_one.py
@@ -773,7 +773,6 @@ def run_import(path, mb_release_id):
 
 
 def _apply_disambiguation(
-    mbid: str,
     album_id: int,
     beets: BeetsDB,
     album_path: str,
@@ -805,10 +804,6 @@ def _apply_disambiguation(
     and ``r.exit_code`` / ``r.decision`` are never touched by a
     disambiguation failure.
     """
-    # ``mbid`` is unused now that ``move_album`` reads the post-move
-    # path by album_id, but the signature is kept as-is so external
-    # callers in tests don't have to re-wire their fixtures.
-    del mbid
     result = move_album(BeetsAlbumHandle(album_id=album_id), beets)
     if not result.success:
         assert result.failure is not None
@@ -934,8 +929,14 @@ def _canonicalize_siblings(
                  "manually later")
             continue
         if result.new_path:
+            # ``move_album`` already ran ``fix_library_modes`` (issue
+            # #84) — log it explicitly so the cratedigger journal trail
+            # still shows per-sibling perm repair. Operators greppable
+            # for 'fix_library_modes' and for 'beet move' can
+            # cross-reference the two in post-hoc debugging.
             _log(f"  [CANONICALIZE] moved sibling id:{aid} → "
-                 f"{result.new_path} (perms repaired)")
+                 f"{result.new_path}; fix_library_modes(...) ran "
+                 "inside move_album")
 
 
 # ---------------------------------------------------------------------------
@@ -1539,7 +1540,7 @@ def main():
     if kept_duplicate:
         _log(f"[DISAMBIGUATE] Running beet move for album id:{pf_info.album_id}")
         album_path = _apply_disambiguation(
-            mbid, pf_info.album_id, beets, album_path, r)
+            pf_info.album_id, beets, album_path, r)
         # Also canonicalize the sibling editions beets flagged as
         # duplicates — without this, the new album gets the ``[YEAR]``
         # suffix while the sibling stays un-suffixed (asymmetric
--- a/harness/import_one.py
+++ b/harness/import_one.py
@@ -48,9 +48,9 @@ from lib.beets_db import AlbumInfo, BeetsDB
 from lib.permissions import fix_library_modes, reset_umask
 from lib.util import beets_subprocess_env
 from lib.quality import (AUDIO_EXTENSIONS_DOTTED as AUDIO_EXTENSIONS,
-                         AudioQualityMeasurement, ImportResult, PostflightInfo,
-                         QualityRankConfig, comparison_format_hint,
-                         determine_verified_lossless,
+                         AudioQualityMeasurement, ImportResult, MovedSibling,
+                         PostflightInfo, QualityRankConfig,
+                         comparison_format_hint, determine_verified_lossless,
                          import_quality_decision, transcode_detection)
 HARNESS = os.path.join(os.path.dirname(__file__), "..", "harness", "run_beets_harness.sh")
 HARNESS_TIMEOUT = 300
@@ -870,7 +870,9 @@ def _remove_stale_by_id_logged(stale_id: int) -> BeetsOpFailure | None:
 
 
 def _canonicalize_siblings(
-    sibling_album_ids: frozenset[int], beets: BeetsDB) -> None:
+    sibling_album_ids: frozenset[int],
+    beets: BeetsDB,
+) -> list[MovedSibling]:
     """Re-run ``beet move`` on each sibling album id after a kept-duplicate import.
 
     When beets' ``%aunique`` disambiguates the new album (adding a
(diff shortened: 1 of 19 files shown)


abl030:
## Codex review verdict: **clean after 5 rounds**

`codex exec review --base main` surfaced real correctness bugs on rounds 1–4, each fixed with a pinning regression test. Round 5 returned clean:

> I did not identify any discrete correctness bugs in the diff relative to main. The new beets-op extraction, release-level advisory locking, and sibling imported_path propagation appear internally consistent and covered by focused tests.

### Findings per round

**R1 P1** — `discogs_albumid` coerced to str in `get_release_ids_by_album_id` (`lib/beets_db.py`). SQLite column is `INTEGER`; `MovedSibling.discogs_albumid: str` + my strict `msgspec.convert` would have raised on every Discogs kept-duplicate import, turning successful beets moves into exception-logged failures. Fixed at the emit site per the .claude/rules/code-quality.md wire-boundary rule. Tests: `TestGetReleaseIdsByAlbumId` × 4. ([56bc6c3](56bc6c3))

**R2 P1** — contention reset used `attempt_type="download"` → `record_attempt` → 30-min exponential backoff. Contention isn't a failed download, so the next cycle was locked out. Dropped the attempt_type; `reset_to_wanted` clears `next_retry_after=NULL`. ([f49572d](f49572d))

**R2 P2** — sibling propagation missed cross-layout Discogs rows. CLAUDE.md documents that Discogs numerics can live in either `mb_release_id` (legacy "pipeline compat") or `discogs_release_id`. The `UPDATE` now matches both columns when `discogs_albumid` is populated. ([f49572d](f49572d))

**R3 P2 + R3 P3** — two more findings on the same contention path (spectral clobber, staging delete). Triggered the fix-bug skill's **stop-and-refactor rule after round 2**: instead of patching another leaf, extracted a real seam. `DispatchOutcome.deferred` now plumbs through the auto-path call chain (`dispatch_import` → `_handle_valid_result` → `_process_beets_validation` → `process_completed_album` → `_run_completed_processing`). Contention becomes a zero-mutation return; the 3-way `bool | None` branch at `_run_completed_processing` replaces the old binary. ([5b5bac6](5b5bac6))

**R4 P1** — lock was acquired INSIDE `dispatch_import_core` but `stage_to_ai` ran BEFORE it. On contention after staging, filesystem state went out of sync with `active_download_state` → next cycle crashed with `FileNotFoundError`. Moved lock acquisition up to `_handle_valid_result`, wrapping `stage_to_ai` AND `dispatch_import`. Reentrant lock semantics mean `dispatch_import_core`'s inner acquire still works. ([2b14b5a](2b14b5a))

**R5** — clean.

### Final test state

- **1976 passing**, 53 skipped.
- Pyright 0 errors on every touched file.
- Codex `exec review` clean.

Ready to deploy via `/deploy`.

Issue #127, opened 2026-04-19
fix(import_one): harden post-import `beet move` disambiguation against TimeoutExpired (#123 follow-up)


abl030:
## Context

PR #126 (closing #123 PR B) hardened \`lib/release_cleanup.py::remove_and_reset_release\` against subprocess failures: each \`sp.run\` is wrapped in try/except covering \`TimeoutExpired\`, \`OSError\`, and non-zero rc, with per-selector failures surfaced via a typed \`ReleaseCleanupResult\`.

The audit for PR B flagged a sibling site with the same structural smell that was explicitly out of scope: \`harness/import_one.py:1185-1204\`, the post-import \`beet move\` that fixes \`%aunique\` path collisions after a duplicate-album import.

## The bug

\`\`\`python
# import_one.py:1185-1204 (approx)
if kept_duplicate:
    move_result = subprocess.run(
        [BEET_BIN, \"move\", f\"mb_albumid:{mbid}\"],
        capture_output=True, text=True, timeout=120,
        env=beets_subprocess_env(),
    )
    if move_result.returncode == 0:
        # Re-read path...
        r.postflight.disambiguated = True
    else:
        _log(f\"  [DISAMBIGUATE] beet move failed (rc={move_result.returncode}): ...\")
\`\`\`

Failure modes:

- **\`TimeoutExpired\` uncaught**: propagates to \`main()\` which logs it and exits code 4 (\"import error\"). But the album WAS already imported to beets. Downstream callers see \`disambiguated: False\` on a malformed \`ImportResult\` JSON and may retry force-import thinking the import failed.
- **\`OSError\` / \`FileNotFoundError\`**: same uncaught propagation.
- **Non-zero rc**: handled but only logged; \`disambiguated=False\` persists as a semi-lie (the import succeeded, just the disambiguation move didn't).

## Suggested shape

Same pattern as PR #126:

1. Wrap the \`subprocess.run\` in a try/except catching \`TimeoutExpired\` and \`OSError\`.
2. Distinguish three outcomes on \`PostflightInfo\`: \`disambiguated=True\` / \`disambiguated=False (not attempted)\` / \`disambiguated=False (move failed: <reason>)\`. The existing \`PostflightInfo\` dataclass can grow a \`disambiguation_error: str | None\` field.
3. Never let this path crash \`import_one.py\` — the album is already on disk successfully; a failed post-import move is a warning, not an import failure.

## Test plan

[...]

PR #143, opened 2026-04-20
refactor: unify wire-boundary encoding on msgspec.Struct (#141)

### LINKED ISSUE #141: refactor: unify wire-boundary encoding on msgspec.Struct
## Motivation

`ImportResult` et al. cross two wire boundaries:
1. Subprocess JSON — `harness/import_one.py` emits `__IMPORT_RESULT__<json>` on stdout; `parse_import_result` decodes.
2. Postgres JSONB — `download_log.import_result` stores it; web API reads re-decode on every request.

The current design is **asymmetric**:

- **Inbound** (decoder): `_postflight_from_dict` uses `msgspec.convert(ms_list, type=list[MovedSibling])` for strict type validation. `msgspec.convert` happily accepts `@dataclass` targets.
- **Outbound** (encoder): `ImportResult.to_json()` uses `json.dumps(asdict(self))`. `dataclasses.asdict` recursively traverses dataclasses but is **opaque to `msgspec.Struct`** (confirmed empirically: it returns the Struct instance unchanged, which then fails `json.dumps`).

So `MovedSibling` is forced to be a `@dataclass` (not a `Struct`) to stay `asdict`-compatible. Meanwhile other wire types in `lib/quality.py` (`HarnessItem`, `HarnessTrackInfo`, `TrackMapping`, `CandidateSummary`, `ChooseMatchMessage`) ARE Structs because they're only decoded (from the harness), never re-encoded through `asdict`.

Problems:

1. **Two rules** for contributors: \"is this type encoded back out?\" → dataclass; \"decoded only?\" → Struct. Easy to get wrong.
2. **Docstrings keep lying.** In PR #136 commit 0110b9c I wrote \"MovedSibling is a msgspec.Struct (wire-boundary type per the rule)\" — it was not, because `asdict` broke. Codex round 5 caught the confusion in final review. The same confusion will hit the next author unless the rule is single-valued.
3. **Strict validation is only inbound.** If we re-encode via `asdict`, nothing validates the encoded shape against the declared types. Future divergence (e.g. a new field added to the type but not to the JSON schema checkers on the web UI) goes undetected.

## Proposal

Commit to one encoder: **`msgspec.json.encode`** (or `msgspec.to_builtins` + `json.dumps` if JSONB writing needs a dict). Every wire-boundary type becomes a `msgspec.Struct`. Single rule: \"types that cross JSON are Structs.\"

In `lib/quality.py::ImportResult.to_json()`:
\`\`\`python
def to_json(self) -> str:
    return msgspec.json.encode(self).decode(\"utf-8\")
\`\`\`

(or similar; `msgspec.json.encode` returns `bytes`, trivially decodable.)

## In scope

Migrate every wire-boundary dataclass to `msgspec.Struct`:

- `ImportResult` (currently `@dataclass`)
- `PostflightInfo`
- `ConversionInfo`
- `SpectralDetail`
- `AudioQualityMeasurement`
- `MovedSibling`
- `BeetsOpFailure` (aliased as `DisambiguationFailure` / `SelectorFailure`)
- `ValidationResult`
- Any others that appear in `ImportResult`'s transitive tree.

Replace every `asdict(...)` on these types with `msgspec.to_builtins(...)` or `msgspec.json.encode(...)`. Remove the asymmetry comments from `_postflight_from_dict` and its neighbors.

Backwards compat: Decoding old JSONB rows must still work. Existing `msgspec.convert` boundary already handles this; add explicit tests for every pre-migration row shape (the ones we've tested historically: pre-#127, pre-#133, pre-current).

## Out of scope

- Non-wire types (`CratediggerConfig`, `DispatchOutcome`, `QualityRankConfig`, `StagedAlbum` from Issue B). These are constructed from typed Python code end-to-end; `@dataclass` is correct for them per `.claude/rules/code-quality.md` § \"Wire-boundary types\".
- Pipeline DB method signatures. They operate on dicts/kwargs, not wire types.

## Acceptance

- Every type that appears in `ImportResult.to_json()`'s output inherits from `msgspec.Struct`.
- Outbound serialisation uses `msgspec.json.encode`, not `json.dumps(asdict(...))`.
- The asymmetry comments in `lib/quality.py::_postflight_from_dict` and `MovedSibling` are removed.
- All pre-existing backwards-compat tests still pass (pre-#127 rows, pre-#133 failure rows, malformed values).
- `.claude/rules/code-quality.md` § \"Wire-boundary types\" updated to say simply: \"Wire-boundary types are `msgspec.Struct`; outbound encoding uses `msgspec.json.encode`.\"
- Pyright 0 errors.

## Risk

- **Positional construction in tests.** Many tests construct `ImportResult(exit_code=0, decision=\"import\", ...)` with specific positional or kwargs. `msgspec.Struct` supports `kw_only=True` — use it. The 100+ test sites that construct these types will need signature audits.
- **Default values + Optional.** `msgspec.Struct` has slightly different default handling than `@dataclass`. Verify `Optional[X] = None` still works.
- **Inheritance.** If any of these types are subclassed, `msgspec.Struct` subclassing rules are stricter than dataclass.

## Process

Use `/refactor` skill. This is the largest refactor of the bunch — touches the wire boundary that every other feature reads through. Expect 4-6 rounds of codex.

Invariant: **\"Wire-boundary types inherit `msgspec.Struct`; encode via `msgspec.json.encode`, decode via `msgspec.convert` — same policy both directions.\"**

Step 3 (probe live system) matters: before changing the encoder, capture a sample of what current production rows serialize as, so the new encoder can be checked for byte-compatibility.

## Context

- PR #136 post-merge reflection, friction point #6.
- `.claude/rules/code-quality.md` § \"Wire-boundary types\" and the issue #99 (PR #98) history that motivated the rule.
- Relevant files: `lib/quality.py`, `lib/beets_album_op.py`, `lib/beets.py`, all test modules that construct these types.

abl030:
## Summary

- Every wire-boundary type in `lib/quality.py` + `lib/beets_album_op.py::BeetsOpFailure` becomes `msgspec.Struct`.
- `ImportResult.to_json` / `ValidationResult.to_json` encode via `msgspec.json.encode` instead of `json.dumps(asdict(self))`.
- Drops the pre-#141 asymmetry (dataclass-if-re-encoded / Struct-if-decoded-only) that consistently produced lying docstrings — one rule both directions.

Closes #141.

## Acceptance (from issue)

- [x] Every type that appears in `ImportResult.to_json()`'s output inherits from `msgspec.Struct`
- [x] Outbound serialisation uses `msgspec.json.encode`, not `json.dumps(asdict(...))`
- [x] Asymmetry comments in `_postflight_from_dict` and `MovedSibling` removed (helper deleted; hedges inlined into `ImportResult.from_dict`)
- [x] All pre-existing backwards-compat tests still pass (pre-#127, pre-#133, v1 rows, malformed values)
- [x] `.claude/rules/code-quality.md` § "Wire-boundary types" updated to the single-policy rule
- [x] Pyright 0 errors (verified via `nix-shell --run "pyright lib/quality.py lib/beets_album_op.py web/routes/pipeline.py lib/import_service.py web/classify.py"`)

## What changed, in order

1. `eb20475` — pin tests for the four production JSONB shapes we must preserve
2. `837ed19` — migration: 8 types → `msgspec.Struct`, unified encoder, asymmetry docstrings gone
3. `0cb75b0` — simplified the wire-boundary rule in `code-quality.md` + CLAUDE.md
4. Five review-fix commits (see the review trail in the PR comment)

## Test plan

- [x] `nix-shell --run "bash scripts/run_tests.sh"` — 2026 tests pass
- [x] `nix-shell --run "pyright <touched files>"` — 0 errors
- [x] Production-shape survey probe on doc2 — every shape in 1726 `download_log.import_result` rows decodes (v1 / current v2 / pre-#127 / pre-#133 / null sub-objects)
- [ ] Post-deploy: grep `/nix/store/*/lib/quality.py` on doc2 for the new `msgspec.json.encode(self).decode()` in `ImportResult.to_json`
- [ ] Post-deploy: verify `cratedigger-web` serves `/api/pipeline/log` without 500s (historical rows still render)
- [ ] Post-deploy: trigger one cratedigger run, verify `ImportResult` JSONB row shape matches what msgspec emits


abl030:
## Review audit trail

Five review rounds; convergence at round 4 (codex clean).

### Round 1 (commit `837ed19` migration)
- **codex P2** → 754119e — legacy callers (`parse_import_result`, `_parse_import_result`, `extract_import_log_fields`) caught `(json.JSONDecodeError, TypeError, KeyError)` but NOT `msgspec.ValidationError`; strict decode of a malformed historical row would now 500 the route / crash a cycle. Extend all three except tuples + pin-test each caller degrades to `None`/`{}`.
- **adversarial** → c397e53 — three MEDIUM/LOW docstring-stale findings plus one MEDIUM pin test for the documented `CandidateSummary` pre-48914ca `mbid` rename trap (rows written before PR #100 silently drop `.mbid=""` instead of raising). The trap itself stays; the pin test locks the contract so a future refactor doesn't accidentally change it.

### Round 2
- **codex P3** → 1416c4e — pre-#141 `_postflight_from_dict` had an `if not d: return PostflightInfo()` guard that tolerated `null` / `[]` / `""` as the `postflight` value. After migration, `msgspec.convert` sees a non-object and raises; the R1 P2 fallback then drops the ENTIRE `ImportResult`, silently losing decision + measurements. Inline a falsy-non-dict hedge in `from_dict`. subTest pin covering the three falsy shapes.
- **adversarial (holistic)** → 4528e6f — the R2 P3 inline hedge left `_postflight_from_dict` orphaned (pyright "is not accessed"). Delete the dead helper + fix four stale references in docstrings/comments that point a future reader at the deleted call site.

### Round 3
- **codex P2** → b4e6943 — documented-equivalence instead of a hedge. Codex flagged hypothetical shapes (`new_measurement: []`, `disambiguation_failure: "string"`) that strict decode would reject. Live probe of 1726 production rows on doc2 shows these shapes DO NOT EXIST — every actual non-object value is JSONB `null`, which msgspec handles natively for `Optional[Struct]`. Per CLAUDE.md "don't add validation for scenarios that can't happen", preserve the strict contract; add a subTest pin asserting every *actually observed* null shape decodes, so a future refactor that forbids `null` on these fields gets caught.

### Round 4 (convergence)
- **codex** → *no findings.*

Reviews disagree productively: adversarial caught docstring lies + dead code (R1 stale comments, R2 orphaned helper), codex caught runtime edge cases (ValidationError-vs-TypeError catch mismatch, falsy-non-dict coercion). The "stop-and-refactor" trigger almost fired at R3 P2 (third round in a row on "strict decode narrower than old tolerance"), but the live-DB probe showed the invariant keeps widening at *theoretical* shapes with zero production incidence — calling that was cheaper than extracting another seam.

PR #150, opened 2026-04-21
[codex] Make library the SSOT for beets and requests

abl030:
## What changed
- make `/api/library/artist` the SSOT view for library management: it now merges beets rows with pipeline-only requests, dedups on exact release identity for both MusicBrainz and Discogs rows, and sorts the merged result once for a stable chronological view
- route the shared delete-from-beets action through `/api/beets/delete` with pipeline context so deleting from the UI also removes the matching pipeline request/history
- purge the pipeline request before the destructive beets delete, so a pipeline-side failure no longer recreates the exact ghost `imported` state this PR fixes
- move ghost-import cleanup into `scripts/cleanup_ghost_imported.py` instead of an inline heredoc in the PR description

## Why
Deleting a release from beets removed the album from the beets DB and disk, but left the `album_requests` row behind at `status='imported'`. That produced a ghost imported state in the UI: the release no longer appeared in the library, but browse rows still showed `imported` and offered `Upgrade` instead of `Add request`.

Separately, the library subview only rendered beets rows, so requests that were not yet on disk had to be managed somewhere else. This change makes that view the SSOT for both owned albums and wanted/requested releases.

## Impact
- deleting from beets now behaves like the pipeline never had that release
- the library subview shows both on-disk albums and pipeline-only requests
- MB and Discogs rows dedup correctly when the same release exists in both beets and the pipeline
- ghost imported cleanup is now a reviewed repo script instead of manual shell quoting

## Validation
- `node tests/test_js_release_actions.mjs`
- `nix-shell --run "python3 -m unittest tests.test_web_server.TestBrowseRouteContracts tests.test_web_server.TestBeetsRouteContracts tests.test_fakes.TestFakePipelineDB tests.test_fakes.TestPipelineDBFakeContract tests.test_beets_db tests.test_cleanup_ghost_imported -v"`
- `nix-shell --run "pyright lib/pipeline_db.py lib/beets_db.py web/routes/browse.py web/routes/library.py scripts/cleanup_ghost_imported.py tests/test_beets_db.py tests/test_cleanup_ghost_imported.py tests/fakes.py tests/test_fakes.py tests/test_web_server.py"`

## Manual Ops
Run from this repo checkout (defaults assume the shared beets DB path and the doc2 pipeline DB DSN):

Dry run:
```bash
nix-shell --run "python3 scripts/cleanup_ghost_imported.py --dry-run"
```

Apply purge:
```bash
nix-shell --run "python3 scripts/cleanup_ghost_imported.py --apply"
```

Override paths/DSN if needed:
```bash
nix-shell --run "python3 scripts/cleanup_ghost_imported.py --dry-run --dsn postgresql://cratedigger@192.168.100.11:5432/cratedigger --beets-db /mnt/virtio/Music/beets-library.db"
```

### DIFF
--- a/lib/beets_db.py
+++ b/lib/beets_db.py
@@ -758,4 +758,5 @@ class BeetsDB:
             "added": r[10], "mb_releasegroupid": r[11],
             "release_group_title": r[12], "min_bitrate": r[13],
             "source": source,
+            "discogs_albumid": str(r[14]) if r[14] is not None else None,
         }
--- a/lib/pipeline_db.py
+++ b/lib/pipeline_db.py
@@ -550,6 +550,8 @@ class PipelineDB:
                 FROM album_requests
                 WHERE mb_artist_id = %s
                    OR (artist_name ILIKE %s ESCAPE '\\'
+                       -- Hyphen-free ids (e.g. legacy numerics / Discogs ids)
+                       -- deliberately fall back to the artist-name match.
                        AND (mb_artist_id IS NULL OR mb_artist_id = ''
                             OR mb_artist_id NOT LIKE '%%-%%'))
                 ORDER BY year, album_title
--- a/lib/pipeline_db.py
+++ b/lib/pipeline_db.py
@@ -28,6 +28,11 @@ DEFAULT_DSN = os.environ.get("PIPELINE_DB_DSN", "postgresql://cratedigger@localh
 BACKOFF_BASE_MINUTES = 30
 BACKOFF_MAX_MINUTES = 60 * 24  # 24 hours
 
+
+def _escape_like_pattern(value: str) -> str:
+    """Escape SQL LIKE wildcards for ``... LIKE %s ESCAPE '\'``."""
+    return value.replace("\\", "\\\\").replace("%", "\\%").replace("_", "\\_")
+
 # Advisory-lock namespaces. Every lock in this codebase is
 # session-scoped, non-blocking (``pg_try_advisory_lock``), and
 # session-reentrant. See ``docs/advisory-locks.md`` for the canonical
@@ -523,6 +528,45 @@ class PipelineDB:
             "SELECT status, COUNT(*) as cnt FROM album_requests GROUP BY status"
         )
         return {r["status"]: r["cnt"] for r in cur.fetchall()}
+
+    def list_requests_by_artist(
+        self,
+        artist_name: str,
+        mb_artist_id: str = "",
+    ) -> list[dict[str, Any]]:
+        """List request rows for one artist, including legacy name fallbacks.
+
+        ``/api/library/artist`` is the SSOT view for albums already in
+        beets and albums still wanted in beets. Prefer exact
+        ``mb_artist_id`` matches when available, but keep the legacy
+        name fallback for older pipeline rows that predate artist-id
+        population or store a non-MB value there.
+        """
+        name_pattern = f"%{_escape_like_pattern(artist_name.strip())}%"
+        if mb_artist_id:
+            cur = self._execute(
+                """
+                SELECT *
+                FROM album_requests
+                WHERE mb_artist_id = %s
+                   OR (artist_name ILIKE %s ESCAPE '\\'
+                       AND (mb_artist_id IS NULL OR mb_artist_id = ''
+                            OR mb_artist_id NOT LIKE '%%-%%'))
+                ORDER BY year, album_title
+                """,
+                (mb_artist_id, name_pattern),
+            )
+        else:
+            cur = self._execute(
+                """
+                SELECT *
+                FROM album_requests
+                WHERE artist_name ILIKE %s ESCAPE '\\'
+                ORDER BY year, album_title
+                """,
+                (name_pattern,),
+            )
+        return [dict(r) for r in cur.fetchall()]
 
     # --- Track management ---
 
--- a/lib/pipeline_db.py
+++ b/lib/pipeline_db.py
@@ -542,6 +542,8 @@ class PipelineDB:
         name fallback for older pipeline rows that predate artist-id
         population or store a non-MB value there.
         """
+        # Pair with `ESCAPE '\'` below so literal `%` / `_` in artist names
+        # do not expand into wildcard matches on PostgreSQL.
         name_pattern = f"%{_escape_like_pattern(artist_name.strip())}%"
         if mb_artist_id:
             cur = self._execute(
--- a/scripts/cleanup_ghost_imported.py
+++ b/scripts/cleanup_ghost_imported.py
@@ -1,0 +1,108 @@
+#!/usr/bin/env python3
+"""Find and optionally purge ghost imported pipeline rows.
+
+A ghost imported row is ``album_requests.status='imported'`` but the
+exact release ID is absent from the beets library. This script uses the
+same exact-ID seam as the web UI (`BeetsDB.locate`) and can either print
+the rows (`--dry-run`, default) or delete them from the pipeline DB
+(`--apply`).
+"""
+
+from __future__ import annotations
+
+import argparse
+import os
+import sys
+from typing import Any
+
+sys.path.insert(0, os.path.join(os.path.dirname(__file__), ".."))
+
+from lib.beets_db import BeetsDB
+from lib.pipeline_db import PipelineDB
+
+DEFAULT_DSN = os.environ.get(
+    "PIPELINE_DB_DSN",
+    "postgresql://cratedigger@192.168.100.11:5432/cratedigger",
+)
+
+
+def _row_release_id(row: dict[str, Any]) -> str:
+    return str(row.get("mb_release_id") or row.get("discogs_release_id") or "").strip()
+
+
+def classify_imported_rows(
+    rows: list[dict[str, Any]],
+    beets: BeetsDB,
+) -> tuple[list[dict[str, Any]], list[dict[str, Any]]]:
+    """Split imported rows into ghosts and manual-review rows."""
+    ghosts: list[dict[str, Any]] = []
+    manual_review: list[dict[str, Any]] = []
+    for row in rows:
+        release_id = _row_release_id(row)
+        if not release_id:
+            manual_review.append(row)
+            continue
+        if beets.locate(release_id).kind == "absent":
+            ghosts.append(row)
+    return ghosts, manual_review
+
+
+def _print_rows(label: str, rows: list[dict[str, Any]]) -> None:
+    print(f"{label}: {len(rows)}")
+    for row in rows:
+        release_id = _row_release_id(row)
+        imported_path = row.get("imported_path") or ""
+        print(
+            f"{row['id']}\t{row['artist_name']} - {row['album_title']}\t"
+            f"{release_id}\t{imported_path}"
+        )
+
+
+def cmd_scan(db: PipelineDB, beets: BeetsDB) -> int:
+    rows = db.get_by_status("imported")
+    ghosts, manual_review = classify_imported_rows(rows, beets)
+    _print_rows("ghost imported rows", ghosts)
+    if manual_review:
+        print()
+        _print_rows("manual review needed (no release id)", manual_review)
+    return 0
+
+
+def cmd_apply(db: PipelineDB, beets: BeetsDB) -> int:
+    rows = db.get_by_status("imported")
+    ghosts, manual_review = classify_imported_rows(rows, beets)
+    for row in ghosts:
+        db.delete_request(int(row["id"]))
+        print(f"deleted {row['id']}: {row['artist_name']} - {row['album_title']}")
+    print(f"\ndeleted {len(ghosts)} ghost imported rows")
+    if manual_review:
+        print()
+        _print_rows("manual review needed (no release id)", manual_review)
+    return 0
+
+
+def main() -> int:
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--dsn", default=DEFAULT_DSN,
+                        help="PostgreSQL DSN for the pipeline DB")
+    parser.add_argument("--beets-db", default="/mnt/virtio/Music/beets-library.db",
+                        help="Path to beets-library.db")
+    mode = parser.add_mutually_exclusive_group()
+    mode.add_argument("--dry-run", action="store_true",
+                      help="Print ghost rows without deleting them (default)")
+    mode.add_argument("--apply", action="store_true",
+                      help="Delete matching ghost imported rows")
+    args = parser.parse_args()
+
+    db = PipelineDB(args.dsn)
+    try:
+        with BeetsDB(args.beets_db) as beets:
+            if args.apply:
+                return cmd_apply(db, beets)
+            return cmd_scan(db, beets)
+    finally:
+        db.close()
+
+
+if __name__ == "__main__":
+    raise SystemExit(main())
--- a/tests/fakes.py
+++ b/tests/fakes.py
@@ -797,6 +797,44 @@ class FakePipelineDB:
             counts[status] = counts.get(status, 0) + 1
         return counts
 
+    def list_requests_by_artist(
+        self,
+        artist_name: str,
+        mb_artist_id: str = "",
+    ) -> list[dict[str, Any]]:
+        needle = artist_name.lower()
+
+        def _legacy_name_match(row: dict[str, Any]) -> bool:
+            artist = str(row.get("artist_name") or "").lower()
+            artist_id = row.get("mb_artist_id")
+            artist_id_str = str(artist_id or "")
+            return (
+                needle in artist
+                and (
+                    artist_id is None
+                    or artist_id_str == ""
+                    or "-" not in artist_id_str
+                )
+            )
+
+        rows: list[dict[str, Any]] = []
+        for row in self._requests.values():
+            if mb_artist_id:
+                if row.get("mb_artist_id") == mb_artist_id or _legacy_name_match(row):
+                    rows.append(copy.deepcopy(row))
+            else:
+                if needle in str(row.get("artist_name") or "").lower():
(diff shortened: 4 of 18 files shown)


abl030:
Addressed the review feedback on this branch:
- moved the library-view pipeline lookup into `PipelineDB.list_requests_by_artist` and added the matching `FakePipelineDB` stub + self-tests
- added the missing delete-route contract fields plus dedup and negative beets-delete coverage
- hardened the shared action toolbar’s inline JS string encoding and updated the PR’s ghost-purge instructions to use writable `psql`

Verification:
- `node tests/test_js_release_actions.mjs`
- `nix-shell --run "python3 -m unittest tests.test_web_server.TestBrowseRouteContracts tests.test_web_server.TestBeetsRouteContracts tests.test_fakes.TestFakePipelineDB tests.test_fakes.TestPipelineDBFakeContract -v"`
- `nix-shell --run "pyright lib/pipeline_db.py web/routes/browse.py tests/fakes.py tests/test_fakes.py tests/test_web_server.py"`

abl030:
## Review — suggestions

### Must-address before merge

**1. Sort-order mixing after merge (`web/routes/browse.py:125-134`)**
Pipeline-only rows are appended *after* beets rows, so the library view will show beets albums (in beets's `added DESC` order) followed by a second block sorted by `(year, album_title)`. Discography views expect a unified chronological timeline. Sort `albums` once after the merge with a deterministic key, e.g. `albums.sort(key=lambda a: (a.get("year") or 0, a.get("album") or ""))`.

**2. Discogs-sourced rows bypass dedup (`web/routes/browse.py:130-133`)**
The dedup key is `mb_release_id`. Pipeline rows for Discogs have `mb_release_id=None` and only `discogs_release_id` populated — `_pipeline_album_rows` maps only `mb_release_id` into `mb_albumid`, so a Discogs request already imported to beets (with `discogs_albumid`) will appear twice. Either add a second dedup path keyed on `discogs_release_id` ↔ beets `discogs_albumid`, or add a test that confirms this double-render is intentional.

**3. Cross-DB failure window in `post_beets_delete`**
Beets delete (SQLite) runs first, pipeline `DELETE` second. If the pipeline delete raises, files are gone but the ghost `imported` row remains — the exact bug this PR fixes. Consider either reversing the order (pipeline first, so a failure leaves a harmless orphan row rather than lost files), or wrapping the pipeline step in a `try/except` that logs loudly and still returns the beets-deletion result.

### Worth fixing while here

**4. Drop `library_rank: \"unknown\"` from pipeline-only rows (`browse.py:97`)**
`badges.js:49` only consults `library_rank` when `in_library=true`. The field is ignored — drop it to avoid the false signal that it does something.

**5. Overlapping queries**
For an album present in both beets and pipeline, `check_pipeline(mbids)` and `list_requests_by_artist(name, mbid)` each hit `album_requests` for the same row. `list_requests_by_artist` could return enrichment info too, eliminating one query.

**6. Legacy name fallback heuristic (`lib/pipeline_db.py:540-548`)**
`mb_artist_id NOT LIKE '%-%'` treats any hyphen-free value as \"not a MB UUID\", which means a Discogs numeric artist id triggers the substring name fallback. Deliberate but fragile — worth a one-line comment explaining this is intentional lenience for pre-MBID rows.

**7. Manual ops cleanup belongs in `scripts/`**
The ghost-imported cleanup in the PR description is a nontrivial inline heredoc with nested quoting. Land it as `scripts/cleanup_ghost_imported.py` so it's reviewable, re-runnable, and doesn't depend on getting SSH quoting right.

### Strengths worth keeping
- `jsArg()` via `JSON.stringify` + HTML-escape is the right fix for the apostrophe injection hole, and the `\"rel-10'oops\"` / `\"12th Man's Greatest Hits\"` regression test is exactly the kind of coverage this needed.
- `_find_pipeline_request_for_release` fallback chain (explicit id → MB release id → numeric Discogs id) is tested end-to-end.
- `FakePipelineDB` contract test for `list_requests_by_artist` keeps the fake honest.
- `_escape_like_pattern` ordering (`\\` before `%`/`_`) is correct.

abl030:
Addressed the second review pass on this branch:
- unified `/api/library/artist` enrichment/dedup around one exact release-identity map, including Discogs rows, and sorted the merged result once after combining beets + pipeline data
- moved the beets delete flow to purge the pipeline row before destructive file removal, with an explicit failure path if the pipeline delete errors
- replaced the PR-body heredoc cleanup with `scripts/cleanup_ghost_imported.py` and added direct tests for the script and the new route behavior

Verification:
- `node tests/test_js_release_actions.mjs`
- `nix-shell --run "python3 -m unittest tests.test_web_server.TestBrowseRouteContracts tests.test_web_server.TestBeetsRouteContracts tests.test_fakes.TestFakePipelineDB tests.test_fakes.TestPipelineDBFakeContract tests.test_beets_db tests.test_cleanup_ghost_imported -v"`
- `nix-shell --run "pyright lib/pipeline_db.py lib/beets_db.py web/routes/browse.py web/routes/library.py scripts/cleanup_ghost_imported.py tests/test_beets_db.py tests/test_cleanup_ghost_imported.py tests/fakes.py tests/test_fakes.py tests/test_web_server.py"`

abl030:
# Review

## Overview

Solid root-cause fix for the ghost `imported` bug plus a UX upgrade that makes `/api/library/artist` the single view for both on-disk albums and pipeline-only requests. Key moves:

- **Delete ordering is now correct**: `/api/beets/delete` purges the pipeline row *first*, aborts on failure, then deletes from beets. Eliminates the failure mode where a pipeline-delete fault recreated the ghost state.
- **Unified library view** merges pipeline rows into `/api/library/artist` with identity-based dedup against beets rows (MBID or numeric Discogs ID).
- **JS string-escape hazard fixed** via new `jsArg()` helper in `release_actions.js` — the old `'${escId}'` pattern broke on apostrophes (e.g. "The 12th Man's Greatest Hits"); there's a regression test for it.
- Inline heredoc cleanup script → reviewed repo script at `scripts/cleanup_ghost_imported.py`.
- `beets_db._album_row_to_dict` now surfaces `discogs_albumid` so Discogs-sourced library rows dedup against Discogs pipeline requests.

## Strengths

- **Correct failure semantics**: pipeline purge happens *before* the destructive beets delete (`web/routes/library.py:142-157`). Exception handler returns 500 and aborts — `beets.delete_album.assert_not_called()` test guards this.
- **Comprehensive test coverage** across all four test layers:
  - Pure: `tests/test_cleanup_ghost_imported.py` classifier with mb vs discogs rows, missing-id manual-review path
  - Seam: `tests/test_beets_db.py` row-to-dict includes `discogs_albumid`
  - Orchestration: 6 new `test_beets_delete_*` tests covering explicit id, release-id fallback, discogs fallback, no-context no-op, pipeline failure aborts, purge=false leaves request
  - Contract: updated `LIBRARY_ALBUM_REQUIRED_FIELDS` and `DELETE_REQUIRED_FIELDS`
- **Shared test infra**: uses `FakePipelineDB`, `make_request_row`, `_WebServerCase` — the new `_configure_beets_delete_mock` helper de-duplicates 6 test bodies.
- **LIKE injection defense**: `_escape_like_pattern` + explicit `ESCAPE '\\'` clause.
- **`jsArg()` helper** is a real win — apostrophe-test in `test_js_release_actions.mjs` is a proper regression guard.

## Issues / Suggestions

### 1. `jsArg()` only used in `release_actions.js`, not `library.js`
`web/js/library.js:271` still uses the old pattern:
```js
onclick=\"...window.confirmDeleteBeets(\${id}, '\${esc(data.artist)}', '\${esc(data.album)}', ...)\"
```
The delete button in `toggleLibDetail` is reachable and will mishandle artist/album names containing apostrophes — exactly the bug the new helper fixes. Per `scope.md` \"clean as you go\" this should have been updated too. Same inconsistency in `library.js:423` (`confirmDeleteBeets` overlay still wraps `releaseId` and `pipelineId` in raw template literals).

### 2. Raw interpolation in `confirmDeleteBeets` overlay
`web/js/library.js:420`: `<p>\${artist} - \${album}<br>\${trackCount} tracks…</p>` inserts raw artist/album into HTML. Callers that use `esc(...)` before invoking end up with double-escaped text displayed to the user (e.g. `Mum &amp; Dad`). Minor pre-existing issue exposed by the new `pipelineNote` path.

### 3. `list_requests_by_artist` substring match is permissive
The `ILIKE %needle%` will match \"Test Artist\" against \"Test Artist Who\" and \"Best Test Artist Ever\". Mitigated by preferring `mb_artist_id` exact match first, but legacy rows without an MBID can drift. Consider tightening to exact match (case-insensitive) unless fuzzy matching is intentional for edit-distance recovery.

### 4. `beets_album_id` falls back to `a.id` — silent type-mixing risk
`web/js/library.js:94`: `const beetsAlbumId = a.beets_album_id || (inLibrary ? a.id : null);`

For pipeline-only rows the backend sets `id = pipeline_id` (`web/routes/browse.py:120`), so if the `in_library` flag ever flips wrong, a pipeline id would be passed as a beets album id downstream. The backend does set `beets_album_id` explicitly on both branches, so the fallback isn't reached — but it's defensive code that hides a backend bug if the contract ever drifts. Consider dropping the fallback and trusting the contract test.

### 5. `detect_release_source` lives in `lib.quality`
Importing `detect_release_source` from `lib.quality` into a web route is fine but surprising — it's a small seam that could live alongside `annotate_in_library` in `lib/artist_compare.py` rather than reaching across into quality.

### 6. `ESCAPE '\\\\'` — fine, but worth a comment
Python-source `'\\\\'` → SQL literal `'\\'` → PostgreSQL escape character `\\` (with `standard_conforming_strings=on`, the default). Works, but a one-line note would save the next reader from the same trace.

## Risks

- **Deployment**: no new migration, pure Python/JS. Safe to roll.
- **Data loss**: pipeline→beets ordering means a successful pipeline purge followed by a beets-delete failure leaves the user without a request but with the beets album. Path exists only if `delete_album` raises `ValueError` *after* the pre-check passed — needs a race. Documented in PR description as intentional.
- **Performance**: `list_requests_by_artist` adds one round-trip per library-artist view. Not cached. Acceptable — artist views are per-click, not background.

## Verdict

Ship-ready once items 1 and 2 are addressed (the `jsArg()` inconsistency especially — the PR introduces the helper explicitly to fix the apostrophe bug, but leaves two reachable call sites still exposed to it). Items 3–6 are polish.

abl030:
Addressed the latest review blockers on this branch:
- moved JS-string argument encoding into shared `web/js/util.js::jsArg()` and wired the remaining `library.js` delete call sites through it
- made the beets-delete confirm overlay escape artist/album itself via a new pure `buildDeleteConfirmHtml()` helper, so the dialog no longer relies on pre-escaped inputs or double-escapes names like `Mum & Dad`
- dropped the `beets_album_id || a.id` fallback in the library renderer and added the requested SQL `ESCAPE` comment in `PipelineDB.list_requests_by_artist`

Verification:
- `node tests/test_js_util.mjs`
- `node tests/test_js_release_actions.mjs`
- `node tests/test_js_library.mjs`
- `node --check web/js/util.js && node --check web/js/release_actions.js && node --check web/js/library.js`
- `nix-shell --run "pyright lib/pipeline_db.py"`

abl030:
Addressed the browse-tab delete-button duplication on this branch:
- extracted a shared `renderRemoveFromBeetsButton()` in `web/js/release_actions.js` and routed both the row toolbar and the library-detail delete button through it
- removed the unused `analysis.js::disambDeleteFromLibrary` path and its `window` export, so there is no second browse-tab beets-delete implementation hanging around
- extended the JS toolbar tests to cover the shared helper with the library-detail styling/onclick shape

Verification:
- `node tests/test_js_release_actions.mjs`
- `node tests/test_js_library.mjs`
- `node --check web/js/release_actions.js && node --check web/js/library.js && node --check web/js/analysis.js && node --check web/js/main.js`

abl030:
# Code Review

## Overview

Five-commit PR (draft). `/api/library/artist` becomes the SSOT merging beets rows + pipeline-only requests with release-identity dedup (MB UUID + Discogs ID). `/api/beets/delete` now purges the matching pipeline request *before* the destructive beets delete, fixing the ghost `imported` state. Adds `scripts/cleanup_ghost_imported.py` for one-off remediation. JS string-embedding hardened via a new `jsArg()` helper. +1205 / −105.

## Strengths

- **Ordering is right.** Pipeline purge *before* `BeetsDB.delete_album` — if the pipeline write fails, the destructive delete aborts (`web/routes/library.py:142-156`). The failure test (`test_beets_delete_pipeline_failure_aborts_before_beets_delete`) asserts `delete_album.assert_not_called()`. Exactly the invariant.
- **Existence check before purge** (`get_album_detail` at library.py:122-128) — avoids orphaning pipeline rows when the beets album id is bogus.
- **LIKE injection hardened.** `_escape_like_pattern` + `ESCAPE '\\'` protects against `%`/`_` in artist names. Hyphen-free `mb_artist_id` fallback (legacy rows / non-UUID values) is well-commented.
- **JS embedding actually fixed.** `jsArg = esc(JSON.stringify(...))` handles apostrophes + HTML-context escaping. Real test proves it (`"Some of the 12th Man's Greatest Hits"`).
- **Shared `renderRemoveFromBeetsButton`** — removed the old duplicated inline-onclick in `library.js` + `analysis.js`. `disambDeleteFromLibrary` is gone (no dead export in `main.js`). Scope rule followed.
- **Test coverage is solid.** Six new `post_beets_delete` orchestration tests cover: explicit id, release-id fallback, discogs-numeric fallback, no-op, failure, no-purge-left-intact. Four library-artist tests cover: merge, MB dedup, Discogs dedup, sort ordering. JS suite adds `test_js_library.mjs`.

## Issues / suggestions

### 1. Pipeline-only rows are missing `library_rank` (contract drift risk)

`LIBRARY_ALBUM_REQUIRED_FIELDS` now includes `library_rank`. The enrichment loop that sets `library_rank` only runs on beets rows (`web/routes/browse.py:155-165`). Rows appended from `_pipeline_album_rows()` never get the key. Nothing breaks today because no contract test asserts required fields on a pipeline-only row — but the route contract guard only catches drift if exercised. Either:

- set `library_rank = None` (or `0`) on pipeline-only rows in `_pipeline_album_rows()`, OR
- extend `test_library_artist_route_includes_pipeline_only_requests` to `_assert_required_fields(..., LIBRARY_ALBUM_REQUIRED_FIELDS, ...)` so the gap is enforced.

Prefer the second — same contract for all rows in the `albums[]` array.

### 2. Orphan risk if `BeetsDB.delete_album` fails *after* pipeline purge

The purge succeeds → `delete_album` raises `OSError`/subprocess failure → the handler returns 500 from the outer `except`, but the pipeline row is already gone. The user sees "error" and has no context left to retry — the ghost state now inverts (row absent, files present).

The test suite guards the pre-delete failure path but not this one. Two options: (a) accept it as a best-effort tradeoff (commit order is deliberate per the PR description) — document it; (b) catch the post-purge exception and emit a targeted error like "files deleted, request purged, stragglers on disk" so the user knows what happened. Worth at minimum a `log.exception` distinguishing the two failure modes.

### 3. `mb_albumid` overloaded with `discogs_albumid` for pipeline-only rows

`web/routes/browse.py:138-143` mutates the beets row to stuff `discogs_albumid` into `mb_albumid` when MB id is absent, then pops `discogs_albumid`. Downstream JS uses `detectSource(mb_albumid)` to disambiguate, so this works — but it's subtle. A one-line comment ("overload mb_albumid for frontend keying; JS routes via detectSource") would save the next reader 10 minutes. Same pattern in `_pipeline_album_rows()` (`mb_albumid: release_id or None`) — deserves the same note.

### 4. Minor: `release_group_title` fallback reuses `album_title`

`_pipeline_album_rows()` sets `"release_group_title": row["album_title"]` so the frontend renders something non-empty, but this is a fib — it's not actually the RG title. Fine for UI, but if anything downstream treats them as equivalent (e.g. dedup by RG title) it will surprise. Nothing in this PR does, but worth flagging as a footgun for future work.

### 5. `refreshAfterBeetsDeletion` has a lot of UI-state conditionals

`web/js/library.js:11-30` dispatches across browse-artist, Recents, Pipeline, and raw DOM removal. The logic is correct but untested. If any of these tabs' markup/labels change (`'Recents'` / `'Pipeline'` text match), this silently drops to the DOM-removal path. Not a blocker, but the tab-name string match would benefit from a data attribute or named constant.

## Risks

- Ghost imported rows can still be produced by external paths (direct `beet remove`, harness-driven drops). The cleanup script mitigates this — keep it as documented ops.
- Discogs release IDs as strings everywhere (`"12856590"`): `_album_row_to_dict` coerces via `str(r[14])`, `_release_identity` strips and compares as str, pipeline side stores as str. Consistent. `BeetsDB.locate` handles both via the exact-ID seam — covered by `test_cleanup_ghost_imported.py`.
- No new pyright errors expected given the validation commands in the PR description.

## Verdict

Solid, scope-appropriate, right-ordered. Worth addressing (1) and adding (2)'s differentiated error before merging — both small. The rest are polish.

abl030:
Addressed the latest review points on this branch:
- pipeline-only `/api/library/artist` rows now carry the same `library_rank` key as beets-backed rows, and the pipeline-only contract test now asserts the full `LIBRARY_ALBUM_REQUIRED_FIELDS` set
- `/api/beets/delete` now distinguishes the post-purge failure mode: if beets deletion fails after the pipeline row was already removed, the route logs that partial-success state and returns a targeted error instead of a generic 500
- added inline breadcrumb comments at the subtle edges the review called out (`mb_albumid` overload for Discogs IDs, placeholder RG title on pipeline rows, uniform row contract, purge-before-delete ordering, and the post-purge failure split)

Verification:
- `nix-shell --run "python3 -m unittest tests.test_web_server.TestBrowseRouteContracts tests.test_web_server.TestBeetsRouteContracts -v"`
- `nix-shell --run "pyright web/routes/browse.py web/routes/library.py tests/test_web_server.py"`

PR #162, opened 2026-04-24
[codex] fix staged-path reflection loose ends

abl030:
## Why this exists

This is a handoff PR that stages the post-PR-156 follow-up work before resetting local state.

The starting point was the post-deploy Claude reflection after PR 156, which called out a handful of "low-severity loose ends" around the staged-path seam:
- add a direct separator matrix test for `staged_filename`
- pin the `db_request_id or 0` fallback behavior at the seam
- add a test for mid-rollback failure in `StagedAlbum.move_to`
- clarify or dedup overlapping repair issue reporting
- narrow the broad `except Exception` in `scripts/repair.py`

Those looked like separate cleanup nits, but they all led back to the same invariant:

> persisted `active_download_state.current_path` must keep a single truthful owner for a download across canonical processing, request-scoped staging, auto-import dispatch, and repair/recovery scans.

## What We Are Actually Solving

The risky cases here are not normal downloads. They are the broken-edge paths where a request has already moved out of the canonical processing dir and the system is deciding whether it is still safe to auto-retry, requeue, mark a failure, or tell repair/manual recovery to stop touching it.

If that seam lies, we either strand files in staging with no audit trail, let repair/poller act as if a staged path is abandoned when it is still live, or reset a request twice after an inner rejection already handled it.

## What Changed

- centralizes path containment in `lib/processing_paths.py` so `download.py` and `download_recovery.py` classify staged roots the same way
- hardens `StagedAlbum.move_to` rollback surfacing and adds direct separator / rollback tests
- routes request auto-import with missing `db_request_id` through the same failed-import + rejection/requeue bookkeeping seam as other auto-import rejects, when the backing request row can be resolved
- keeps missing-id request-scoped staged paths blocked for manual recovery instead of guessing ownership
- switches repair's auto-import liveness probe to a read-only `pg_locks` check instead of transiently acquiring the release lock itself
- clarifies blocked-path reporting when a request-scoped staged path has no MBID, because auto-import cannot resume at all in that case
- replaces an impossible repair dedupe test with direct tests for dedupe and lock-probe behavior
- adds an explicit slice proving `process_completed_album() -> False` does not cause a second transition after an inner rejection already requeued the row

## Why The Weeds Got This Deep

This follow-up wandered because each "small" reflection item exposed another place where the staged-path seam was partially duplicated or partially unaudited. The next reviewer should review this as one seam cleanup, not as ten unrelated test additions.

The real question for review is:

> when a request is partway between canonical processing, request-scoped staging, and auto-import, does every caller still agree about who owns the path and what recovery actions are safe?

## Reviewer Focus

Please focus on:
- whether `_resolved_request_rejection_id()` is the right defensive way to recover the backing row when `db_request_id` is missing
- whether the `pg_locks` probe in `scripts/repair.py` matches the release-lock shape we actually hold in production
- whether the changed repair wording and no-MBID handling reflect the intended operator story
- whether the false-outcome / no-double-transition contract is now pinned tightly enough

## Validation

- `nix-shell --run "python3 -m unittest tests.test_staged_album tests.test_download tests.test_download_recovery tests.test_repair_cli tests.test_integration_slices -v"`
- `nix-shell --run "pyright lib/download.py lib/download_recovery.py lib/processing_paths.py scripts/repair.py tests/test_download.py tests/test_download_recovery.py tests/test_repair_cli.py tests/test_integration_slices.py"`
- `nix-shell --run "python3 -m unittest tests.test_integration_slices.TestRunCompletedProcessingOutcomeBranching -v"`

All green locally.

## Reviewer Note

I started another partner-review pass while packaging this and then stopped because the goal here is to checkpoint the work before resetting state. Treat this PR as the handoff artifact for the next review round.

### DIFF
--- a/lib/download.py
+++ b/lib/download.py
@@ -615,10 +615,11 @@ def process_completed_album(album_data: GrabListEntry, failed_grab: list[Any],
       dispatch succeeded.
     - ``False`` — a non-deferred failure happened. Outer caller resets to
       ``wanted`` only if the request row is still ``downloading``.
-    - ``None`` — processing left the row untouched. This covers both
-      release-lock contention and post-move staged paths that must not
-      be auto-retried because they already live under
-      ``beets_staging_dir``. Outer caller must NOT touch status.
+    - ``None`` — processing left the row untouched. This covers release-lock
+      contention, post-move staged paths that must not be auto-retried
+      because they already live under ``beets_staging_dir``, and guarded
+      request rejects where ownership cannot be proven. Outer caller must
+      NOT touch status.
     """
     staged_album = StagedAlbum.from_entry(
         album_data,
@@ -669,8 +670,10 @@ def _process_beets_validation(album_data: GrabListEntry,
     Returns the dispatch outcome when the auto-import path fires,
     ``None`` when beets validation rejects (``_handle_rejected_result``
     already handles the state transition) or when the non-auto
-    redownload path takes over in ``_handle_valid_result``. Caller
-    uses the ``deferred`` flag to decide whether to flip status.
+    redownload path takes over in ``_handle_valid_result``. Guarded
+    ownership-less rejects also return a deferred outcome so callers
+    keep the row untouched for manual recovery. Caller uses the
+    ``deferred`` flag to decide whether to flip status.
     """
     from lib.beets import beets_validate as _bv
     from lib.preimport import run_preimport_gates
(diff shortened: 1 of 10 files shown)


abl030:
Addressed the final staged-path review gaps on this branch:
- repair now ignores missing request-scoped auto-import paths while the release lock shows import is still running, and the lock probe now matches the correct advisory-lock shape
- request auto-import rejects now fail closed when no safe request row can be proven, while the safe year/MBID recovery paths and the single-transition reject path are pinned by tests

Verification:
- nix-shell --run "python3 -m unittest tests.test_download tests.test_download_recovery tests.test_repair_cli tests.test_integration_slices -v"
- nix-shell --run "pyright lib/download.py lib/download_recovery.py scripts/repair.py tests/test_download.py tests/test_download_recovery.py tests/test_repair_cli.py tests/test_integration_slices.py"

Issue #82, opened 2026-04-12
Discogs: no cover art for Discogs-only releases


abl030:
## Context

The Discogs CC0 data dump does not include images. MusicBrainz has the Cover Art Archive (coverartarchive.org) which provides freely available cover art by release ID. Discogs-only releases therefore have no cover art in the browse UI or library views.

## Options to explore

1. **Cross-reference to MusicBrainz CAA** — for releases that exist in both Discogs and MB, look up the MB release ID and fetch from coverartarchive.org. Many albums overlap between the two databases.

2. **Discogs API image proxy** — the live Discogs API (not the dump) provides image URLs, but requires authentication and is rate-limited (60 req/min). Could proxy through a cache layer. More complex and depends on an external service.

3. **User-uploaded covers** — allow manual cover art upload for Discogs releases. Most invasive, but gives full control.

4. **Accept the limitation** — cover art is cosmetic. The pipeline (search, download, validate, import) works without it. Beets' fetchart plugin may find covers from other sources during import anyway.

## Shipped in

PR #79 (issue #69) — Discogs as first-class pipeline citizen.

abl030:
Closing as moot (triage 2026-07-02):

- The web UI renders no cover art for **any** source — no `<img>` tags, no coverartarchive references anywhere in `web/`. MB releases don't get CAA art in browse/library views either, so this was never a Discogs-specific gap; the UI is deliberately text-only and has worked fine that way since April.
- Option 4 ("accept the limitation") is what de facto happened, and its prediction held: beets runs `fetchart` + `embedart` (both `auto: yes`) on every import, so library files get on-disk + embedded art regardless of source. Meelo and Plex — the actual listening surfaces — display art from the files.
- Option 2 (live Discogs API proxy) would add an authenticated, rate-limited external dependency for cosmetics; option 3 (user uploads) is exactly the machinery the single-operator scope rules exist to kill.

If web-UI cover art ever becomes genuinely wanted, that's a new feature covering both sources (CAA for MB + something for Discogs) — file fresh against that actual desire.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #22, opened 2026-04-05
Rename spectral DB columns for clarity


abl030:
## Problem

`spectral_bitrate` vs `on_disk_spectral_bitrate` — the difference between "download's spectral" and "current disk state" is only visible if you already know the architecture. This contributed to the field mixup in #18.

## Proposal

Rename to make the distinction self-documenting:
- `spectral_bitrate` → `last_download_spectral_bitrate`
- `spectral_grade` → `last_download_spectral_grade`
- `on_disk_spectral_bitrate` → `current_spectral_bitrate`
- `on_disk_spectral_grade` → `current_spectral_grade`

Low priority — would require a migration + updating all readers/writers. But would prevent the class of bug where someone reads the wrong one.

## Context

Post-fix reflection from #18.

Issue #21, opened 2026-04-05
Type spectral measurements to prevent field confusion


abl030:
## Problem

`DownloadInfo` is a 20+ field grab-bag where `spectral_bitrate` means "the download's spectral" in some contexts and "what import_one measured" in others. The quality gate bug in #18 was partly caused by `_check_quality_gate()` reading `spectral_bitrate` (download) when it meant `on_disk_spectral_bitrate` (disk state) — the field names are too similar and carry no type distinction.

## Proposal

A `SpectralMeasurement` dataclass pairing `grade` + `bitrate` — always set together, never one without the other. Then `DownloadInfo` would carry:
- `download_spectral: SpectralMeasurement | None` — what spectral found on the new files
- `on_disk_spectral: SpectralMeasurement | None` — what's currently on disk

The type system catches "you read the download's spectral but meant the on-disk spectral" at pyright time, not at runtime in production.

## Context

Post-fix reflection from #18. The field mixup would have been a type error under this design.

abl030:
Implemented in PR #23 (merged to main, deployed to doc2).

`SpectralMeasurement` frozen dataclass in `lib/quality.py` with `grade` and `bitrate_kbps` — always paired, never one without the other. `from_parts()` factory returns None when both fields are None.

`DownloadInfo` now carries:
- `download_spectral: SpectralMeasurement | None` — what spectral found on the new files
- `current_spectral: SpectralMeasurement | None` — what's currently on disk

`GrabListEntry` follows the same pattern, replacing 4 loose fields with 3 typed ones.

The #18 field mixup (`spectral_bitrate` vs `on_disk_spectral_bitrate`) is now a type-level distinction that pyright catches — you can't accidentally read the download measurement when you meant the on-disk measurement.

Issue #20, opened 2026-04-05
Consolidate spectral state writes into single function


abl030:
## Problem

`on_disk_spectral_bitrate`, `on_disk_spectral_grade`, `spectral_bitrate`, and `spectral_grade` are written piecemeal across 3 files with independent guard conditions:

1. `download.py:349` — pre-import spectral gather (`if is not None`)
2. `album_source.py:236` — `mark_done()` post-import (`if spectral_grade`)
3. `import_dispatch.py:117` — quality gate reads for decisions

These four fields always change together — they represent "what spectral analysis found." But each writer evolved independently with its own guard, leading to #18 where one field updated but another stayed stale.

## Proposal

A single function like `update_spectral_state(db, request_id, grade, bitrate)` that always writes all four fields (including NULL). The guard moves to the call site ("did spectral run?") rather than being duplicated inside each write.

This makes it structurally impossible for one field to go stale while another updates.

## Context

Post-fix reflection from #18. The `mark_done()` guard has been fixed, but the pattern remains in `download.py:349` (latent, not currently causing bugs).

abl030:
Implemented in PR #23 (merged to main, deployed to doc2).

`RequestSpectralStateUpdate` in `lib/pipeline_db.py` — a frozen dataclass with `last_download` and `current` fields (both `SpectralMeasurement | None`). `as_update_fields()` expands to the correct column names. `update_spectral_state()` method on PipelineDB delegates to it.

All spectral writes now go through this typed path:
- `mark_done()` in `album_source.py` — post-import
- `_apply_spectral_decision()` in `download.py` — pre-import
- `_apply_request_spectral_fields()` in `import_service.py` — force-import/manual-import

The latent inconsistency from #18 (`download.py:349` using `if is not None` guard) is fixed — it now goes through `update_spectral_state()` which always writes both grade and bitrate together, including explicit NULLs.

PR #185, opened 2026-04-29
feat: Label viewer Phase A (Discogs-first)

### LINKED ISSUE #183: Label viewer
I.E. search for record labels jsut like we do arists and then have the same album/ep split view over time. but this will lead to hundreds of results, so perhaps another way to filter, year?

abl030:
Implements the label viewer for cratedigger ([#183](https://github.com/abl030/cratedigger/issues/183)) per `docs/plans/2026-04-29-001-feat-label-viewer-phase-a-plan.md`. Search labels, browse a label's catalogue with album/EP split + library overlay + year/format filters, drill-in from any release row.

Phase A is Discogs-only with source-agnostic plumbing so Phase B (MusicBrainz labels) is a wiring task.

## Scope

Two repositories:

- **discogs-api** ([abl030/discogs-api@07f7e5e](https://github.com/abl030/discogs-api/commit/07f7e5e)) — three new Rust endpoints + GIN FTS index. Already deployed.
- **cratedigger** (this PR) — adapter, routes, UI, drill-in.

## Implementation units (matches plan U-IDs)

| Unit | Commit | Summary |
|------|--------|---------|
| U1+U2 | discogs-api `07f7e5e` | `/api/labels` search, `/api/labels/{id}`, `/api/labels/{id}/releases` with recursive-CTE sub-label rollup |
| U3 | `767d354` | `web/discogs.py`: `search_labels`, `get_label`, `get_label_releases` + source-agnostic `LabelEntity` (`msgspec.Struct`) decoded at the wire boundary |
| U4 | `f45739b` | `web/routes/labels.py` + library/pipeline overlay; extracted shared helper `web/routes/_overlay.py` and refactored two call sites in `browse.py` to use it |
| U5+U6 | `1316a1d` | `web/js/labels.js`: search UI as a third toggle in browse tab + label-detail page with `renderTypedSections` reuse, year/format/hide-held filters as pure JS predicates |
| U7 | `55a9039` | Clickable label links in release-detail panel; `renderLabelLinks` helper in `web/js/labels.js` (escapes user-controlled text, falls back to plain text for MB releases without IDs) |

## Key decisions

- **Source-agnostic from day one.** `LabelEntity` shape is shared across MB and Discogs adapters; nullable fields (country, parent_label_id) accommodate both. Phase B is a wiring task.
- **Sub-label rollup on by default.** `?include_sublabels=false` opt-out. Frontend defaults to `false` for labels with `release_count > 1000` (UMG-class) to dodge the recursive-CTE perf cliff.
- **`/api/discogs/label/*` route shape** mirrors existing `/api/discogs/artist/*` convention (per-source URL prefix). Phase B will add `/api/label/{mbid}` parallel to `/api/artist/{mbid}`.
- **`browse.py` overlay refactor**: extracted near-duplicate inline overlay logic from `get_release_group` and `get_discogs_master` into `web/routes/_overlay.py::overlay_release_rows`. Identical behavior preserved; covered by existing tests.
- **Library overlay is the load-bearing filter** (per brainstorm). Year + format are secondary refinements that compose on top.
- **No labelmate-recommendation sort** (Approach D dropped from brainstorm). No editorial label metadata, no cover art (Discogs CC0 has none — #82).

## Testing

- **Python:** 2,487 tests pass. New `TestLabelRouteContracts` adds 6 contract tests including the integration scenario (one release in beets + one in pipeline DB → both correctly overlaid). `TestRouteContractAudit` registers the two new patterns. Pyright clean inside nix-shell.
- **JS:** 122/122 pure-function tests pass (`node tests/test_js_util.mjs`). New coverage on `applyLabelFilters`, `parseYear`, `sortByYearDesc`, `buildLabelSearchUrl`, `renderLabelLinks` (including XSS guard).
- **Wire-boundary regression:** msgspec.ValidationError test feeds `release_count` as a string and asserts the boundary catches it.
- **Live smoke test:** `curl https://discogs.ablz.au/api/labels?name=hymen+records` returns Hymen Records (id=265, 659 releases) — endpoints deployed and working before this PR.

## Post-Deploy Monitoring & Validation

After merging this PR and bumping the cratedigger flake input on doc1:

- **Validation window:** First 30 min after `nixos-rebuild switch` on doc2.
- **Owner:** abl030 (manual playwright walkthrough).
- **Healthy signals:**
  - `ssh doc2 'curl -s "http://127.0.0.1:8080/api/discogs/label/search?q=hymen"'` returns JSON with non-empty `results`
  - `curl https://music.ablz.au/api/discogs/label/265` returns label + 100 releases with overlay populated
  - Browser walkthrough of AE1 (Gridlock release → click "Hymen Records" → label page renders with library overlay correct) succeeds
  - `cratedigger-web.service` starts cleanly: `ssh doc2 'sudo journalctl -u cratedigger-web --since "5 min ago"'`
- **Failure signals & rollback:**
  - 500s on `/api/discogs/label/*` → check `journalctl -u cratedigger-web` for adapter exceptions; likely msgspec.ValidationError on Discogs payload drift → rollback flake input
  - Artist view broken (overlay missing) → `_overlay.py` extraction regressed something not caught by tests; rollback
  - JS console errors on browse tab → likely `state.js` schema drift; rollback
- **Log searches:** `journalctl -u cratedigger-web | grep -E '(/api/discogs/label|labels\.py|overlay_release_rows)'`

## Known follow-ups (deliberate, not blockers)

- **FTS index activation.** `idx_label_name_fts` on `discogs-api` takes effect on the next monthly `discogs-import` run. Until then, label search uses sequential scan over 2.36M label rows (slow but functional). Trigger a manual import if needed.
- **Big-label recursive CTE.** UMG-scale labels (>2.7M releases via tree) take 36s with `include_sublabels=true`. Frontend mitigation in place (auto-flips to `false` when `release_count > 1000`); follow-up: add covering composite index `release_label(label_id, release_id)` if needed.
- **Pagination forwarding.** Route doesn't yet forward `?page` / `?per_page` to the adapter (which already supports them). First page (100 rows) is enough for Phase A; deep pagination is a small follow-up if any label exposes the gap in practice.
- **Phase B (MusicBrainz labels).** Separate plan. Plumbing is ready (`LabelEntity` source-tagged shape, route convention parallel to artist view).

## Closes

Closes #183.

### DIFF
--- a/docs/brainstorms/label-viewer-requirements.md
+++ b/docs/brainstorms/label-viewer-requirements.md
@@ -1,0 +1,135 @@
+---
+date: 2026-04-29
+topic: label-viewer
+---
+
+# Label Viewer
+
+## Problem Frame
+
+When the user finds a release they like — e.g. Gridlock on Hymen Records — they want to see what else came out on that label, and quickly tell which of those they already own and which they haven't heard. Today the cratedigger web UI surfaces labels only as plain-text fields on Discogs release payloads (`web/discogs.py:231`, `web/discogs.py:282`); there is no way to navigate from a release to the rest of its label, no way to search by label name, and no library overlay to mark "already mine". The workaround is leaving cratedigger and using Discogs.com directly, which loses the library overlay that makes the local mirror valuable in the first place.
+
+This is fundamentally a discovery surface, not a navigation surface — the goal is "what I haven't heard but probably want to", not "let me look up Warp's catalogue". That framing drives the filter priorities below.
+
+Origin: GitHub issue #183.
+
+---
+
+## Key Flows
+
+- F1. Drill-in from a release
+  - **Trigger:** User is viewing a release in the browse tab or library and clicks the label name.
+  - **Steps:** Label page opens with the label's name, optional metadata (country, parent label if applicable), and the full release list. Library overlay marks each release as `in library` / `in pipeline` / `not held`. Year and format filters available; sort newest-first by default.
+  - **Outcome:** User can scan the label end-to-end and instantly see which releases are new to them.
+  - **Covered by:** R1, R2, R5, R6, R7, R9, R10
+
+- F2. Top-down label search
+  - **Trigger:** User types a label name (e.g. "Hymen") into a search box in the browse tab.
+  - **Steps:** Search returns a disambiguation list of matching label entities with enough context to pick the right one (release count, country, parent label where present). User clicks a result to land on the same label page from F1.
+  - **Outcome:** User reaches the label page without first needing to find a release on it.
+  - **Covered by:** R3, R4, R5
+
+- F3. Era-bounded browse on a large label
+  - **Trigger:** User is on a label page for a high-volume label (Warp, Blue Note) where the flat list is unwieldy.
+  - **Steps:** User applies year filter and/or format filter (album / EP / single, optionally physical format) to narrow the list. Library overlay still applies.
+  - **Outcome:** User sees a scoped, scannable list for that era / format.
+  - **Covered by:** R6, R7, R9
+
+---
+
+## Requirements
+
+**Search and navigation**
+- R1. A release detail view that today shows label names as plain text MUST expose each label as a clickable link to that label's page.
+- R2. The label page MUST be reachable directly by URL (deep-linkable, shareable, refresh-safe).
+- R3. The browse tab MUST offer label search alongside the existing artist search, returning a disambiguation list of matching label entities.
+- R4. Label search results MUST include enough context to disambiguate similarly-named labels: at minimum release count and country; parent label name shown when the result is a sub-label.
+
+**Label page content**
+- R5. The label page MUST list every release on the label that the data source knows about, paginated or virtualized for labels with hundreds or thousands of entries.
+- R6. Each release row MUST show release title, primary artist, year, and format/type, and link to the existing release detail page.
+- R7. The label page MUST split releases by primary type the same way the artist discography view does (album / EP / single / other), preserving the existing visual conventions.
+
+**Library overlay (primary discovery filter)**
+- R8. Each release on the label page MUST be marked with its current library status: `in beets library`, `in pipeline (any non-imported state)`, or `not held`. The marking MUST be visually scannable, not require hover.
+- R9. The label page MUST allow filtering to "not held" releases as a single-click toggle, since this is the dominant discovery move.
+
+**Secondary filters**
+- R10. The label page MUST provide year and format/type filters that compose with the library overlay (all three layered).
+- R11. Default sort MUST be year descending (newest first); within the same year, secondary sort is stable but unspecified.
+
+**Source-agnostic plumbing (enables Phase B without rework)**
+- R12. The label entity model exposed to the UI MUST be source-tagged (e.g. `{source: "discogs" | "musicbrainz", id, name, ...}`), not Discogs-shaped, even though only Discogs is wired in v1.
+- R13. The label page route MUST accept a source-qualified identifier, so adding MusicBrainz later is a new source adapter behind the same route shape rather than a parallel route.
+- R14. The release-list rendering on the label page MUST work from a source-agnostic release shape compatible with the artist discography view's existing release shape, so the same component renders MB and Discogs label release lists when MB is added.
+- R15. The "add MusicBrainz label support" follow-up MUST NOT require schema, route, or component changes to anything shipped in v1 — only a new source adapter implementation and the search-side wiring.
+- R16. The source-agnostic shape MUST be defined to accommodate what the upstream MusicBrainz API actually returns, not what we'd prefer it to return. The MusicBrainz mirror runs the upstream MusicBrainz Server codebase and we cannot modify its API surface; the Discogs mirror is ours and we can extend its endpoints freely. This asymmetry MUST be reflected in the v1 design: any normalization or denormalization required to hit the source-agnostic shape happens in our Python web layer (per-source adapter), not server-side.
+
+---
+
+## Acceptance Examples
+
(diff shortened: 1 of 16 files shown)


abl030:
## Code review pass complete

Ran `/ce-code-review plan:docs/plans/2026-04-29-001-feat-label-viewer-phase-a-plan.md` (Tier 2, 4 parallel reviewer subagents covering correctness, adversarial, security, api-contract, reliability, performance, maintainability, testing, project-standards, kieran-python, julik-frontend-races). 23 surviving findings after dedup + cross-reviewer promotion + 75-anchor confidence gate. Run artifacts at `/tmp/compound-engineering/ce-code-review/20260429-ded45010/`.

**Verdict before fixes:** Ready with fixes. **After fixes:** Ready to merge — the P0 and full safe_auto cluster have been addressed.

### Fixed in `bf2d929` (`fix(web): auto-flip include_sublabels on big labels and tighten label-detail validation`) and `a1cffb0` (`refactor(web): label adapter cleanup and JS hardening`)

| # | Finding | Resolution |
|---|---------|------------|
| **P0 #1** | DoS via UMG-class recursive CTE — single `curl /api/discogs/label/<umg-id>` froze `music.ablz.au` for 36s | Two-layer fix: (a) Rust mirror sets Postgres `statement_timeout = '5s'` on the recursive CTE and returns 503 on timeout; (b) cratedigger route auto-flips `include_sublabels=false` when `release_count > 1000` unless caller passed the param explicitly. New constant `BIG_LABEL_THRESHOLD = 1000` in `web/routes/labels.py` cross-referenced with `web/js/labels.js`. 2 new contract tests. |
| **P2 #5** | discogs-api `include_sublabels=false` ORDER BY incorrect (DISTINCT ON wrap broke pagination) | Wrapped DISTINCT ON in subquery, outer `ORDER BY released, id` |
| **P2 #6** | Big-label fallback dead code (3 reviewers caught this) | *Deferred — blocked by `loadLabelReleases` throw semantics; addressing in follow-up alongside the upstream-503 handling note below.* |
| **P2 #9** | `distinctFormats` had no unit tests | 6 table-driven cases covering empty/dedup/multi-value/whitespace/null format rows |
| **P3 #10** | Year filter coerced NaN to truthy, silently dropping the filter | Hardened at both `onLabelFilterChange` (input parsing) and `applyLabelFilters` (defense in depth); 4 NaN unit tests |
| **P3 #11** | `searchLabels` swallowed fetch errors → outer catch dead | Removed try/catch swallow; throws on `!r.ok` so caller's catch fires correctly |
| **P3 #12** | `?include_sublabels=` accepted garbage strings as truthy | Strict `{true,false,1,0}` (case-insensitive); 400 otherwise; 2 contract tests |
| **P3 #13** | `get_label_releases` not wrapped in HTTPError 404 catch | Mid-flight 404 between `get_label` and `get_label_releases` now surfaces as 404 not 500; regression test added |
| **P3 #14** | Cache round-trip used hand-rolled `_label_entity_to_dict`/`_from_dict` | Replaced with `msgspec.to_builtins(entity)` and `msgspec.convert(d, type=LabelEntity)`; helpers deleted |
| **P3 #15** | `search_labels` cache key omitted `per_page` | Added `page`/`per_page` keyword-only params; both included in cache key |
| **P3 #21** | Dead import: `loadReleaseGroup` | Removed; `node --check` passes |
| **P3 #22** | `Optional[T]` vs `T \| None` style inconsistency | Normalized the new label structs + `LabelEntity` to `T \| None`; dropped unused `Optional` import |

### Tests
- Python: **2487 → 2493** tests pass (+6 new contract tests for label-detail).
- JS: **122 → 134** assertions pass (+12 covering `distinctFormats` and `applyLabelFilters` NaN cases).
- pyright: 0 errors / 0 warnings inside nix-shell.

### Residual findings — left for follow-up

These are concrete `gated_auto` / `manual` items the reviewer team flagged but the user opted to surface rather than auto-apply this round:

- **P1 #2** Pagination plumbed but not consumed — first 100 only. Plan U6 explicitly accepted this for v1; recommend a follow-up issue.
- **P1 #3** Recursive CTE `UNION ALL` → cycle in `parent_label_id` would infinite-loop. Discogs probably has no label cycles in practice; defensive switch to `UNION` or PG14+ `CYCLE` clause.
- **P1 #4** Rust struct fields typed `String` for nullable `label.profile`/`contactinfo`/`data_quality`. Schema has `NOT NULL DEFAULT ''` so unlikely to trigger; defensive `Option<String>` + `unwrap_or_default()`.
- **P2 #7** Total-count header is direct-only count but list rolls up sub-labels — visible UI mismatch for big labels.
- **P2 #8** `openLabelDetail` has no in-flight token: rapid clicks on different labels could land out of order.
- **P3 #16** `overlay_release_rows` mutation contract is implicit; cache-poisoning risk if a future caller passes a cached dict.
- **P3 #17** `overlay_release_rows` lacks a direct unit test (only indirect via `get_release_group` and `get_discogs_master`).
- **P3 #18** Hardcoded `sub_labels: []` in label-detail response — discogs-api returns sub_labels but route drops them. Either pass through (UI doesn't render them today) or remove the field.
- **P3 #19** Year inputs fire on every keystroke without debounce. Negligible at 100 rows; will matter when pagination lands.
- **P3 #20** Adapter accepts arbitrary `int|str` `label_id` and interpolates raw into URL (defence-in-depth — route regex already restricts at entry).
- **P3 #23** Rust `via_label_id` field naming differs from rest-of-schema `label_id`.

### Discovered during the fix pass — new follow-up

- **Upstream 503 handling.** The Rust mirror now returns 503 when the recursive CTE exceeds 5s. The cratedigger route doesn't have a special handler for 503; `urllib` raises `HTTPError(503)` which surfaces as 500 to the browser. The auto-flip should prevent this in the happy path, but a user explicitly passing `?include_sublabels=true` on a UMG-class label would still see a 500. Consider a graceful retry-without-sublabels on 503 in a follow-up.

### Pre-existing (informational, not blocking this PR)

- Stale label-search results can overwrite fresh ones — same race exists in the existing artist search.
- Cache key for label search uses unbounded user query (no length cap) — same pattern as existing `search_artists`.

### Discogs-api side

The corresponding statement_timeout + ORDER BY fix is on `discogs-api@cf22645` (already pushed). Deploy needs a `discogs-src` flake input bump on doc1 + `nixos-rebuild switch` on doc2 — same flow as the original U1+U2 deploy.

Issue #101, opened 2026-04-18
Web cache: split MB metadata cache from pipeline-status overlay


abl030:
## Background

Today the web UI caches \`/api/release-group/<id>\`, \`/api/release/<id>\`, \`/api/discogs/master/<id>\`, and \`/api/discogs/release/<id>\` Redis-side. These responses bake \`pipeline_status\` and \`in_library\` into each pressing alongside the MB/Discogs metadata.

That mix causes a stale-badge bug: soularr-the-pipeline updates \`album_requests.status\` (downloading→imported→wanted on quality-gate re-queue) directly in Postgres. Those writes happen outside the web UI's POST-mutation path, so the cached payload's \`pipeline_status\` doesn't get invalidated. With the original \`TTL_MB=86400\` (24h), users saw "downloading" badges for hours after an album was actually re-queued.

## Tactical fix already shipped (commit on main)

Dropped the affected endpoints from \`TTL_MB\` (24h) to \`TTL_LIBRARY\` (5min). Bounded staleness at 5 minutes, no architectural change. See \`web/server.py::Handler._CACHE_TTLS\`.

## Architectural fix this issue tracks

Stop baking the pipeline-status overlay into the MB/Discogs metadata cache. Two reasonable directions:

### Option A — separate cache layers
- Cache pure MB/Discogs metadata at TTL_MB (24h).
- Layer pipeline_status + in_library on every request via two cheap queries (\`check_pipeline\` + \`check_beets_library\`, both already exist).
- The overlay is fast — single SQL per album list — so paying for it on every request is fine.
- Requires reshaping the cache key to exclude per-pressing pipeline data, plus a render-time merge.

### Option B — pubsub invalidation
- soularr-the-pipeline publishes a Redis pubsub event on every \`album_requests.status\` write.
- Web server subscribes and invalidates \`web:/api/release*\` and \`web:/api/discogs/{master,release}*\` keys when the affected MBID/Discogs ID is hit.
- Smaller code change but introduces a runtime coupling (web depends on pubsub being reliable; missed events = stale).

Option A is the architecturally cleaner one — eliminates the staleness class entirely instead of papering over it. Option B has nicer best-case behaviour (cache stays warm longer when nothing changes) but worse worst-case (missed pubsub = silent stale).

## When this becomes worth doing

[...]

PR #440, opened 2026-06-12
test(web): #430 batch 1 — FakePipelineDB harness mode, 4 modules migrated, ratchet baseline

abl030:
First batch of #430 — migrating the `tests/web` contract harness off the MagicMock pipeline DB onto the stateful `FakePipelineDB`, with a ratchet that makes the migration self-enforcing.

## Why

Both adversarial review passes on 2026-06-12 (#428's P1, #429's testing gap) hit the same root cause: `tests/web/_harness.py` injects a MagicMock DB, so contract tests pass whenever the mock's shape matches the assertion's shape — production query semantics never enter the loop. This is the `.claude/rules/test-fidelity.md` meta-pattern, structurally embedded in the harness.

## What

1. **Ratchet** (`tests/_mock_audit_scanner.py` + `tests/test_mock_audit.py`): `WEB_HARNESS_MOCK_BASELINE` pins per-file occurrence counts of `mock_db` references (tests/web) and `_pipeline_db_test_harness` references (everywhere). Exact-match: growth fails the audit, shrink forces the baseline entry to drop in the same commit. **The migration is done when the dict is empty.** Counter semantics pinned by a subTest table covering the alias/bare-reference evasion shapes.
2. **`_FakeDbWebServerCase`** (`tests/web/_harness.py`): per-test fresh bare `FakePipelineDB` installed as `web.server.db` (the production DSN-less module-global swap). Tests seed state (`seed_request`, `log_download`, `update_status`) and assert against the fake's real query semantics.
3. **Four modules migrated** (baseline entries deleted): `test_routes_youtube.py`, `test_server_cache.py`, `test_routes_browse.py`, `test_routes_library.py`. Notable upgrades in fidelity:
   - the #101 stale-badge regression now flips status via `db.update_status(42, "downloading")` — the actual "pipeline writes directly to Postgres" sequence
   - library search/recent now **prove** pipeline enrichment runs: the legacy MagicMock cursor silently iterated empty so `web.overlay.check_pipeline`'s raw-SQL path never executed in tests (the `check_pipeline` patch in those tests was dead weight); they now queue a `FakeCursor` row projected from the seeded request and assert the 900 kbps pipeline floor lands as `min_bitrate=900000`
   - purge-failure ordering pinned via a typed `_FailingDeleteDB(FakePipelineDB)` subclass instead of `side_effect`
4. **Shared `FakeCursor`** in `tests/fakes.py` (promoted from `test_web_overlay.py`'s private `_Cursor`, with self-tests): psycopg2 consumption parity — `fetchone` advances, `fetchall` drains, `while cur.fetchone():` terminates. `FakePipelineDB._execute` unqueued default is now an empty cursor ("query ran, zero rows") instead of a None that 500s opaquely.

## Remaining (subsequent PRs)

Baseline: `_harness.py` 39, `test_routes_imports.py` 71, `test_routes_pipeline.py` 66, `test_routes_pipeline_mutations.py` 100, `test_routes_pipeline_replace.py` 27, `test_routes_pipeline_search_plan.py` 42, `test_server_endpoints.py` 53. The legacy `_WebServerCase` mock defaults and `_pipeline_db_test_harness` die with the last entry.

**Known residual risks** (documented, deliberate):
- A new tests/web file can still subclass legacy `_WebServerCase` and lean on `_make_server`'s class-level mock defaults with zero counted occurrences — closed only when the legacy harness is deleted at the end of the series.
- Per-test `patch.object(srv, "db", ...)` lifetime vs background work: theoretical for these synchronous routes, but the `test_routes_imports.py` migration (TriageRunner sweeps) must account for handler threads outliving a test.

## Acceptance checklist

- [x] Full suite green (4402 tests, `OK`), including the three gates (JS, dead-code, unittest)
- [x] `nix-shell --run pyright` — 0 errors full-repo
- [x] Ratchet RED-proven both directions (growth fail + shrink fail observed during development)
- [x] Migrated modules contain zero `mock_db` / `_pipeline_db_test_harness` occurrences
- [x] Same-engine adversarial pre-review: 1 HIGH + 2 MEDIUM found, all fixed with pinning tests (see review-fix commit)
- [x] Codex partner review (`codex exec review --base main`): no findings

Closes nothing yet — #430 stays open until the baseline dict is empty.

### DIFF
--- a/tests/_mock_audit_scanner.py
+++ b/tests/_mock_audit_scanner.py
@@ -522,7 +522,6 @@ _WEB_HARNESS_OVERRIDE_RE = re.compile(
 WEB_HARNESS_MOCK_BASELINE: Dict[str, int] = {
     os.path.join("web", "_harness.py"): 31,
     os.path.join("web", "test_routes_imports.py"): 63,
-    os.path.join("web", "test_routes_library.py"): 4,
     os.path.join("web", "test_routes_pipeline.py"): 59,
     os.path.join("web", "test_routes_pipeline_mutations.py"): 98,
     os.path.join("web", "test_routes_pipeline_replace.py"): 26,
--- a/tests/_mock_audit_scanner.py
+++ b/tests/_mock_audit_scanner.py
@@ -521,15 +521,12 @@ _WEB_HARNESS_OVERRIDE_RE = re.compile(
 
 WEB_HARNESS_MOCK_BASELINE: Dict[str, int] = {
     os.path.join("web", "_harness.py"): 31,
-    os.path.join("web", "test_routes_browse.py"): 4,
     os.path.join("web", "test_routes_imports.py"): 63,
     os.path.join("web", "test_routes_library.py"): 4,
     os.path.join("web", "test_routes_pipeline.py"): 59,
     os.path.join("web", "test_routes_pipeline_mutations.py"): 98,
     os.path.join("web", "test_routes_pipeline_replace.py"): 26,
     os.path.join("web", "test_routes_pipeline_search_plan.py"): 42,
-    os.path.join("web", "test_routes_youtube.py"): 1,
-    os.path.join("web", "test_server_cache.py"): 3,
     os.path.join("web", "test_server_endpoints.py"): 51,
 }
 
--- a/tests/_mock_audit_scanner.py
+++ b/tests/_mock_audit_scanner.py
@@ -490,3 +490,62 @@ def scan_tree() -> Dict[str, Dict[str, int]]:
         if counts:
             result[rel] = counts
     return result
+
+
+# === Web-harness MagicMock ratchet (#430) ===
+#
+# The tests/web contract harness historically injected a MagicMock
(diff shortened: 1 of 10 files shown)


abl030:
## Review audit trail

| Round | Reviewer | Findings | Resolution |
|-------|----------|----------|------------|
| r1 | Same-engine adversarial (Claude, minimal context) | 1 HIGH (ratchet evasion via bare/aliased `mock_db` refs + line-vs-occurrence counting), 2 MEDIUM (`FakeCursor.fetchone` non-consuming; `_execute` None default → opaque 500), 2 LOW (state contradiction in overlay cursor row; `_pipeline_db_test_harness` invisible outside tests/web), 2 NIT (docstring half-truth; two idioms in library module) | All fixed in `review(r1 adversarial)` commit, each with a pinning test or by-construction equivalence. One LOW (stalled-handler timing window) documented as residual risk for the imports-module PR rather than fixed — no migrated route does post-response DB work. |
| r2 | Codex partner (`codex exec review --base main`) | none | — |

Convergence verdict: zero unfixed correctness findings; residual risks documented in the PR body.

PR #619, opened 2026-07-11
feat(triage): surface unreferenced quarantine folders

abl030:
## Summary

- add one typed read-only service that compares immediate `failed_imports/` album directories with the live Wrong Matches projection
- expose identical results through `pipeline-cli triage quarantine [--json]` and `GET /api/triage/quarantine`, including route descriptions and self-discovery
- normalize legacy relative, absolute, and descendant `failed_path` references against the configured slskd download root
- exclude code-owned `bad_files` and `untracked_audio` buckets, ignore files/symlinks, avoid recursion, and make output deterministic with name/path/mtime metadata
- fail the complete view closed as CLI 5 / HTTP 503 on config, DB, decode, scan, stat, or mid-scan race errors; a genuinely absent quarantine root remains a valid empty result
- document the operator workflow and lifecycle/safety contract

This implements **W2 only** of #573. It does not close the covering issue and adds no deletion or automatic decision path.

## Invariant

The surfaced set is exactly the immediate real album folders under the configured quarantine root that have no visible Wrong Matches reference. Relative, absolute, and descendant references protect their containing immediate album root; references outside the configured quarantine do not claim local folders; special category roots never surface or recurse. Any incomplete observation is unavailable, never a partial orphan list.

## TDD and verification

- RED: service, generated, CLI, and API contracts initially failed because `lib.quarantine_triage_service` did not exist
- RED race pin: an entry disappearing after `scandir` opened was initially misreported as empty; narrowed missing-root handling and made mid-scan disappearance unavailable
- deterministic pin + generated property + known-bad invariant-checker self-test shipped together
- new generated module fuzz profile: 20k examples, OK
- targeted service/CLI/API/route/docs/mock/dead-code audits: OK
- full suite: JS gates + vulture + 5,234 Python tests, OK (129.703s)
- whole-repo pyright: 0 errors
- signed commit verified locally
- pre-push: all randomized generated-test shards + `nix flake check` (moduleVm, eval guards, CLI bundle), OK


abl030:
Reviewer P1 fixed in signed commit `70303beaf85ca5158c0de39d0df4a96b6030663c`.

- Centralized default Wrong Matches visibility in `lib.wrong_matches.wrong_match_row_is_visible`.
- Both the default card builder and quarantine reference collector now exclude frozen `request_status='replaced'` rows through that same predicate.
- `include_replaced=true` still opts the explicit history view back in through the shared predicate.
- RED deterministic reproduction: replaced-only folder incorrectly returned `folders=[]` before the fix.
- RED generated shrink: `[('relative', 'replaced')]`.
- Generated world now ranges across request status and path representation; the existing known-bad invariant-checker qualification remains.

Verification after the fix: focused 100-test surface green; quarantine fuzz profile 20k examples green; full suite 5,230 tests green; whole-repo pyright 0 errors; pre-push randomized generated shards and `nix flake check`/moduleVm green.

abl030:
Implemented the second review round in signed commit `5ec852c3a06b212d3664ff09901cbf3be894ef41`.

Changes:
- `pipeline-cli` now parses the command before unrelated mirror setup, so `triage quarantine` does not depend on mirror configuration.
- Runtime-config and `PipelineDB` constructor failures on the quarantine path now return the stable exit-5 JSON/human contract; successful handles retain the existing `finally` close behavior.
- Other commands retain mirror-base configuration and their prior exception behavior.
- The web route now maps DB acquisition failures to the stable quarantine 503 response, while unexpected service bugs still surface as generic 500s.
- Added real entrypoint/HTTP-harness regression tests, including DB close and non-quarantine preservation coverage.

RED evidence before the fix:
- CLI DB-constructor failures escaped in both JSON and human modes.
- Web DB acquisition returned the generic 500 response.
- The runtime-config failure was outside the quarantine error boundary.

Verification:
- focused entrypoint/service/web/audit tests: 75 passed
- full suite: 5,242 passed
- whole-project pyright: 0 errors
- quarantine generated fuzz: 20,000 examples per property
- pre-push randomized generated suite and `nix flake check`: passed

Issue #1214, opened 2026-08-19
Daily gate ENOSPC: BeetsContractWorld leaks one tmpfs world per Hypothesis example (~13 G of the 14 G fuzz peak)


abl030:
## Symptom

`cratedigger-daily-checks.service` failed 2026-08-20 05:05→05:28 on doc1:

```
generated fuzz burst: INFRASTRUCTURE ABORT after 495.0s
  (1368 completed of 4586; 3218 not started; property verdict invalid)
OSError: [Errno 28] No space left on device
DiskFull: could not extend file "base/16384/156217": No space left on device
Consumed 4h 38min 15s CPU over 23min 49s wall, 26.4G memory peak, 414M swap peak
```

The deterministic suite in the same run PASSED. The unit gets its own private
16 GiB tmpfs (`TemporaryFileSystem=/run/cratedigger-daily-checks/scratch:rw,size=16G`,
`nixosconfig/modules/nixos/ci/cratedigger-daily-checks.nix:401-404`); the fuzz
burst filled it.

## Root cause

`BeetsContractWorld` is constructed **inside `@given` bodies** with cleanup
registered via `unittest`'s `addCleanup`:

```python
# tests/test_beets_config_contract_generated.py:156-157
world = BeetsContractWorld()
self.addCleanup(world.close)
```

Hypothesis re-executes the test body once per example; `addCleanup` fires once
per **test method**. Every example therefore leaks a live world until the method
finally returns. Each world allocates *two* `tempfile.TemporaryDirectory` trees
(`tests/fakes/beets_contract.py:280-286`): one under `TMPDIR` and one under
`/dev/shm`.

**40 construction sites**: 24 in `tests/test_beets_config_contract_generated.py`,
16 in `tests/test_beets_config_contract_regressions_generated.py`.

### Measured

One test class at the daily's exact per-shard budget
(`CRATEDIGGER_FUZZ_MAX_EXAMPLES=2500`), sampled every 2 s:

```
max_tmpdir_worlds=2469   max_bytes=1576.99MB   max_shm_worlds=2450
```

**1.58 GB resident from a single test method**, growing monotonically to the end
of the method, then released at once. Burst-level confirmation: running only
`tests.test_beets_config_contract_generated` at `JOBS=8`, 2500 examples, for 85 s
drove tmpfs to **2.73 GB**, with the entire top-8 consumer list being
`beets-contract-*` at 638,976 bytes each.

[...]

PR #1239, opened 2026-08-21
feat(tests): bound suite runs to their own cgroup (#1238)

### LINKED ISSUE #1238: Suite runs are unbounded: a runaway phase takes the whole host, not just itself
## Symptom

On 2026-08-19 the #1214 leak drove doc1 into a **global** OOM:

```
oom-kill:constraint=CONSTRAINT_NONE,...,global_oom
```

`CONSTRAINT_NONE` means no cgroup limit bounded the allocation, so the kernel picked victims across the whole machine by `oom_score` rather than killing the offender. Six processes died — every one a Python `MainThread` worker holding 0.7–3.0 GB — and four gate units ended `Failed with result 'oom-kill'`.

## Why it still matters after #1214

The leak is fixed. The *shape* is not: nothing limits a suite run to its own share of the host, so the next runaway again takes out whatever else the operator had running. That collateral includes the Claude daemon, which shares this user slice and owns the operator's background agent sessions — a suite bug silently becomes lost agent work.

Every existing guard (`TEST_RAM_ROOT_EXHAUSTED`, `TEST_HOST_MEMORY_EXHAUSTED`, `_check_suite_headroom`) **classifies** an exhaustion that already happened, or refuses admission before one. None **bounds** a run in progress.

## Measured baseline

`scripts/daily_resource_monitor.sh` telemetry from 2026-08-20 (post-fix, healthy):

| phase | peak user-slice memory | samples |
| --- | --- | --- |
| `deterministic_suite` | 22.41 GB | 15,746 |
| baseline (same run) | 14.56 GB | — |

So the suite's own share is roughly 8 GB, against a 32 GiB host.

## Proposed

Launch the suite inside a transient `systemd-run --user --scope` carrying a `MemoryMax` derived from the host's own MemTotal (never a constant — #1079), so a runaway is reclaimed and killed inside its own cgroup and the host survives.

Two non-obvious constraints found while measuring:

- **`MemorySwapMax=0` is required.** `MemoryMax` alone bounds only resident memory; on a host with swap the overage spills instead of being refused. Measured: 200 MiB under a 64 MiB cap exits 0 without it, 137 with it.
- **`MemoryHigh` must NOT be set.** It stalls a runaway in the reclaim loop rather than killing it — measured at over 120s with nothing to reclaim — turning an unattended gate into a hang.

The two nightly runners are launched by nixosconfig **system** units with no reliable user D-Bus, so they need a declarative unit-level `MemoryMax=` instead; wiring the helper there would fail open on exactly the unattended path that failed.

abl030:
Closes #1238.

## What

A suite run had no memory limit of its own. The 2026-08-19 #1214 leak therefore produced a **global** OOM — `oom-kill:constraint=CONSTRAINT_NONE,...,global_oom` — so the kernel chose victims across the whole host by `oom_score` and killed six unrelated processes rather than the offender. The leak is fixed; the shape is not. The next runaway's collateral includes the Claude daemon in the same user slice, which owns the operator's background agent sessions.

Every existing guard (`TEST_RAM_ROOT_EXHAUSTED`, `TEST_HOST_MEMORY_EXHAUSTED`, `_check_suite_headroom`) classifies an exhaustion after the fact, or refuses admission before one. None bounds a run in progress.

`scripts/memory_scope.sh` emits a transient `systemd-run --user --scope` prefix. `scripts/test.sh` and `scripts/run_final_gate.sh` launch through it.

## Sizing is measured, not chosen

`MemoryMax` = 70% of the host's own MemTotal — derived, never a constant, so other installations scale (#1079). Margin comes from `daily_resource_monitor.sh` telemetry, 2026-08-20 (post-#1214-fix, healthy):

| phase | peak user-slice memory | samples |
| --- | --- | --- |
| `deterministic_suite` | 22.41 GB | 15,746 |
| baseline, same run | 14.56 GB | — |

Suite's own share ≈ 8 GB; the cap on this 32 GiB host is ~22.4 GiB, about 2.8× headroom.

## Two findings that came from measuring, not reasoning

Both were wrong in my first draft, and the file records both.

**`MemorySwapMax=0` is load-bearing.** `MemoryMax` alone bounds only *resident* memory; on a host with swap the overage spills instead of being refused, so a runaway thrashes rather than dying. Measured:

| config | 200 MiB alloc under 64 MiB cap |
| --- | --- |
| `MemoryMax` only | **exit 0** — not contained |
| `+ MemorySwapMax=0` | **exit 137** — killed |

**`MemoryHigh` must not be set.** It looks like the obvious gentler first tier and is a trap: with swap capped it stalled a runaway in the kernel's reclaim loop for **over 120 seconds** with nothing to reclaim, and had to be killed by hand. For an unattended nightly gate that converts "fail fast with 137" into "hang forever and never report" — worse than the uncontained behaviour this fixes.

## Fail-open vs fail-closed

Asymmetric, deliberately:

- **Missing infrastructure** (no `systemd-run`, no delegated memory controller, unreadable MemTotal) → warn, run uncontained. Containment is a safety net, not a correctness gate; refusing to run the suite would trade a rare host OOM for a common hard stop.
- **Invalid explicit override** → fail closed. An operator who typo'd a limit must not silently get a different one. Mirrors `headroom_floor_bytes`.

`CRATEDIGGER_TEST_MEMORY_MAX_BYTES` overrides the limit; `0` disables containment.

*The operator did not pick between these when asked; this is my recommendation, and it is one line in `cratedigger_memory_scope_prefix` to flip.*

## Nightly runners deliberately opt out

`daily_flake_update.sh` / `daily_beets_tip_update.sh` run under nixosconfig **system** units (`Type=oneshot`, `User=abl030`) with no reliable user D-Bus, so `systemd-run --user` would fail open to no containment on exactly the unattended path that failed. They get a declarative unit-level `MemoryMax=` in nixosconfig instead — **not in this PR**. Both files document the opt-out, and a test asserts that documentation exists, so an undocumented absence can't pass for an oversight.

## Testing

`tests/test_memory_scope.py`, 17 deterministic tests. Test infrastructure, so deterministic-only per `code-quality.md`.

These pin the **argv the helper emits**, not the kernel's enforcement of it — whether `MemoryMax` kills a process is a property of cgroup v2, and driving a real scope from the suite would make the whole suite depend on a live user D-Bus session. Enforcement evidence is the measured table above.

Each fail-open branch is driven to its own condition with earlier branches passing, and asserts its own distinct message, so a short-circuit can't let one assertion stand in for another.

### Kill matrix

| Site | Mutant | Test that goes RED | Result |
| --- | --- | --- | --- |
| `-p MemorySwapMax=0` | delete the flag | `test_prefix_caps_swap` | KILLED |
| `_CRATEDIGGER_MEMORY_MAX_PERCENT=70` | → `60` | `test_default_limit_is_seventy_percent_of_mem_total` | KILLED |
| absence of `MemoryHigh` | re-add `-p MemoryHigh=$((max/2))` | `test_does_not_set_memory_high` | KILLED |
| invalid-override `return 1` | → `return 0` | `test_invalid_override_fails_closed` | KILLED |
| `!= *" memory "*` | → `!= *memory*` | `test_substring_controller_name_is_not_a_match` | KILLED |
| nesting guard | condition → `false` | `test_already_scoped_run_does_not_nest` | KILLED |
| `test.sh` prefix application | drop the argv expansion | `test_every_suite_launcher_sources_and_applies_the_prefix` | KILLED |

Control (unmutated tree) green; 7/7 killed; tree verified reverted after the run.

### Gates

`bash scripts/test.sh tests.test_memory_scope` → **6/6 phases pass** (js-syntax, js-unit, pyright, ruff, vulture, python). That run also exercised the new wiring for real — `test.sh` now launches through the scope — and emitted zero `uncontained` warnings, so containment was established live.

## Docs

`.claude/rules/code-quality.md` gains the knob and the two constraints, in the section that already owns `CRATEDIGGER_TEST_RAM_MIN_BYTES` / `CRATEDIGGER_TEST_MEMORY_MIN_BYTES`. Registry entries added for `scripts/memory_scope.sh` and `scripts/run_final_gate.sh` (which had none), and `tests.test_memory_scope` added to `scripts/test.sh`'s.

### DIFF
--- a/.claude/rules/code-quality.md
+++ b/.claude/rules/code-quality.md
@@ -617,6 +617,30 @@ rather than promoting to infrastructure-failure — a known, accepted
 residual (independent review F10), not a claim that the two never
 co-occur.
 
+Every guard above CLASSIFIES an exhaustion that already happened; none
+BOUNDS it. Until #1238 nothing limited a suite run to its own share of the
+host, so the 2026-08-19 leak produced a `CONSTRAINT_NONE ... global_oom` —
+the kernel choosing victims across the whole machine by `oom_score`, killing
+six unrelated processes rather than the offender. `scripts/memory_scope.sh`
+closes that: `scripts/test.sh` and `scripts/run_final_gate.sh` launch the
+suite inside a transient `systemd-run --user --scope` carrying
+`MemoryMax` (70% of the host's own MemTotal — derived, never a constant, so
+other installations scale) plus `MemorySwapMax=0`. The swap cap is
+load-bearing, not belt-and-braces: `MemoryMax` alone bounds only RESIDENT
+memory, so on a host with swap a runaway spills over and thrashes instead
+of dying (measured: 200 MiB under a 64 MiB cap exits 0 without it, 137
+with it). `MemoryHigh` is deliberately absent — it stalls a runaway in the
+reclaim loop for minutes rather than killing it, turning an unattended gate
+into a hang. `CRATEDIGGER_TEST_MEMORY_MAX_BYTES` overrides the limit; `0`
+disables containment. Missing infrastructure (no `systemd-run`, no delegated
+memory controller) warns and runs UNCONTAINED — containment is a safety net,
+not a correctness gate — but an invalid explicit override fails closed, the
+same asymmetry `headroom_floor_bytes` already uses. The two nightly runners
+deliberately opt out and are bounded by a declarative `MemoryMax=` on their
+nixosconfig system units instead: `systemd-run --user` needs a user D-Bus
+session those units do not reliably have, so wiring the helper there would
+fail open on precisely the unattended path that failed in the first place.
+
 **Generated (property-based) production tests**
 (`tests/test_*_generated.py`, Hypothesis)
 run deterministically in the suite. After changing quality policy, run the
--- a/scripts/daily_beets_tip_update.sh
+++ b/scripts/daily_beets_tip_update.sh
@@ -39,6 +39,14 @@ nix flake update beets-tip mutagen-tip mediafile-tip
 # entry-time free-bytes refusal defers to it (issue #1111). Every automated
 # launcher of the canonical suite sets this.
 export CRATEDIGGER_SUITE_OWNS_HEADROOM=1
+# No scripts/memory_scope.sh containment here, deliberately: this runner is
+# launched by the nixosconfig `cratedigger-beets-tip-canary` SYSTEM unit
+# (Type=oneshot, User=abl030), which has no reliable user D-Bus session, so
+# `systemd-run --user` would fail-open to no containment on exactly the
+# unattended path that most needs it. The nightly units carry a declarative
+# `MemoryMax=` on the unit itself instead -- enforced by the system manager,
+# no bus required. The helper is for the interactive/agent launchers
+# (scripts/test.sh, scripts/run_final_gate.sh) that do have a user session.
 nix develop .#tip --command bash scripts/run_tests.sh
 
 # Deliberately no lock commit or push. The canary's product is the signal,
--- a/scripts/daily_flake_update.sh
+++ b/scripts/daily_flake_update.sh
@@ -129,6 +129,15 @@ run_stage() {
     daily_resource_monitor_set_phase runner_overhead
 }
 
+# No scripts/memory_scope.sh containment in this runner, deliberately. It is
+# launched by the nixosconfig `cratedigger-daily-checks` SYSTEM unit
+# (Type=oneshot, User=abl030), which has no reliable user D-Bus session, so
+# `systemd-run --user` would fail open to no containment on exactly the
+# unattended path that suffered the 2026-08-19 global OOM. These stages are
+# bounded by a declarative `MemoryMax=` on the unit itself, enforced by the
+# system manager with no bus involved. `stable_nix` would be unbounded either
+# way: `nix build` delegates the work to the nix-daemon, which runs in the
+# system slice under its own cgroup, outside anything this script controls.
 daily_resource_monitor_set_phase runner_overhead
 run_stage deterministic_suite "deterministic full suite" \
     env CRATEDIGGER_SUITE_OWNS_HEADROOM=1 \
--- a/scripts/memory_scope.sh
+++ b/scripts/memory_scope.sh
@@ -1,0 +1,174 @@
+#!/usr/bin/env bash
+# Transient-cgroup memory containment for canonical suite launches.
+#
+# WHY THIS EXISTS
+#
+# On 2026-08-19 a leaking test phase (#1214, one tmpfs world per Hypothesis
+# example) drove doc1 into a GLOBAL out-of-memory event. The kernel log is
+# explicit about the blast radius:
+#
+#     oom-kill:constraint=CONSTRAINT_NONE,...,global_oom
+#
+# CONSTRAINT_NONE means no cgroup limit bounded the allocation, so the kernel
+# chose victims across the WHOLE machine by oom_score rather than killing the
+# offender. Six processes died, every one a Python `MainThread` worker holding
+# 0.7-3.0 GB, and four gate units ended `Failed with result 'oom-kill'`.
+#
+# The leak itself is fixed. What is NOT fixed is the shape: nothing bounds a
+# suite run to its own share of the host, so the next runaway again takes out
+# whatever else the operator happened to be running -- including the Claude
+# daemon that owns their background agent sessions, which shares this user
+# slice and would be picked off as collateral.
+#
+# A transient scope with a memory limit converts that global event into a
+# local one: reclaim pressure applies to the suite's own cgroup first, and
+# past the hard limit only the suite's processes are killed. The host, and
+# everything else on it, survives.
+#
+# SIZING IS DERIVED, NEVER A CONSTANT
+#
+# A limit tuned to this homelab's 32 GiB box is precisely the
+# host-specific rule this repository rejects (#1079), and `scope.md` requires
+# that defaults not assume this installation. Both limits are therefore
+# fractions of the host's own MemTotal.
+#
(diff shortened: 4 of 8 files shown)


abl030:
## Independent review: blocking. Do not merge as written.

I verified the review's findings myself rather than accepting them. Four are real and blocking. The review was wrong on one detail — my 22.41 GB / 15,746-sample figure does reproduce exactly from `/run/user/1000/cratedigger-daily-resource.8ciEBe/samples.tsv` — but that turns out to make the sizing error worse, not better, because it means I measured the wrong cgroup.

### 1. The cap is sized against the wrong cgroup and would kill healthy runs

`daily_resource_monitor.sh:433-443` resolves its cgroup from `/proc/self/cgroup` — under the nightly `cratedigger-daily-checks` **system unit** that cgroup starts near-empty, so the phase peak *is* the gate's own footprint. The file I measured instead had a 14.56 GB floor at `bootstrap`, i.e. a user-session run already holding Claude and everything else. My "subtract the baseline, so the suite is ~8 GB" inference is invalid for the path this protects.

The authoritative nightly receipts:

| date | status | peak | peak owner |
| --- | --- | --- | --- |
| Aug 17 | valid | 21.05 GB | `generated_fuzz` |
| Aug 18 | valid | 19.89 GB | `generated_fuzz` |
| Aug 19 | valid | 21.41 GB | `generated_fuzz` |
| Aug 20 | **invalid** (`monitor_failed`, ENOSPC) | — | — |
| Aug 21 | valid | 17.72 GB | `deterministic_suite` |

Cap is 70% of 32.02 GiB MemTotal = **24.06 GB** against a measured worst case of **21.41 GB** → **1.12× headroom**, not the 2.8× claimed. Aug 19 also used 3.2 GB of swap, which `MemorySwapMax=0` forces back into RAM. A healthy nightly would be at real risk of being killed by its own containment.

Note also the peak owner is `generated_fuzz` on 3 of 4 valid runs — a phase none of the wired launchers covers.

### 2. Fail-open is not implemented for the failure mode that actually occurs

The precondition checks cgroup *delegation*; `systemd-run --user` needs a reachable *user bus*. Verified with `XDG_RUNTIME_DIR` pointed at a bus-less directory:

```
helper_rc=0 argc=10          # full prefix emitted, no warning
systemd-run ... -> rc=1      # Failed to connect to user scope bus
```

So it fails **closed**, hard. `test.sh` execs into exit 1; `run_final_gate.sh` gets status 1, `record_suite_bundle` fails, and it writes `fail 1` to the receipt — **`status` then reports `fail` for a suite that never ran**, a state unreachable from a bus problem before this PR. This also falsifies the "would fail open" rationale I repeated in five places, including the justification for excluding the nightly runners.

### 3. Containment blinds the existing OOM classifier

`TEST_HOST_MEMORY_EXHAUSTED` requires `BrokenProcessPool` **and** system-wide `MemAvailable < 256 MiB`. A cgroup kill guarantees system memory is healthy, so it can never fire on the contained path — the failure degrades into N ordinary markers named after innocent tests, exactly the N-disguises shape #1111/#1156 exist to prevent. `TEST_RAM_ROOT_EXHAUSTED` doesn't catch it either (verified: a capped scope writing to the shared tmpfs gets SIGKILL, not ENOSPC).

### 4. Three surviving mutants at the launcher-wiring tests

Both launcher tests are source-substring greps, so no functional property is pinned:

| mutant | result |
| --- | --- |
| `test.sh` computes the prefix, prints it to `/dev/null`, execs uncontained | **SURVIVED** 17/17 |
| `run_final_gate.sh`: `\|\| die` → `\|\| true` (no longer fails closed) | **SURVIVED** 32/32 |
| delete the `source` line, keep the `# shellcheck source=` comment, add a no-op stub | **SURVIVED** 17/17 |

The first is precisely the defect the feature exists to prevent, surviving a test named `test_every_suite_launcher_sources_and_applies_the_prefix`. My kill-matrix row for that site was true only of one literal-text mutant.

### 5. False founding claim

"the kernel chose victims across the whole host … rather than the offender" / "killed six unrelated processes" is wrong. All six victims *were* the suite's own workers (`cratedigger-gate-1204*.service` ×4, `tmux-spawn-*` ×2) — `oom_score` picked the actual hogs. Bystander protection is a legitimate *prospective* motive; it is not what was measured. The helper header contradicts itself in the same paragraph.

### Also

- Six separate OOM events 13–23 min apart, victims 0.69–3.17 GB each — aggregate pressure across runs, not one runaway. A 24 GB per-run cap binds none of them individually.
- `MemTotal` drifts on this Proxmox guest (34918371328 → 34813513728 within one session; `Balloon:4914936kB` at the time of the incident), so a MemTotal fraction is an unstable basis here.
- `scripts/test.sh:33` sources *after* `cd`, breaking `bash repo/scripts/test.sh` from a parent dir and `cd scripts && bash test.sh` — both worked before this PR.

### Verified correct, keep

- `MemorySwapMax=0` is load-bearing — reproduced: 200 MiB under a 64 MiB cap is exit 0 without it, exit 137 with it.
- `run_final_gate.sh`'s `$!` / `proc_start_ticks` identity still holds: `systemd-run --scope` **execs** the target, so `$!` is the target's own PID.
- `--setenv` applies to the scope, so the nesting guard works.

### Where this leaves the design

The suite legitimately peaks at ~21 GB on a 32 GB box. A per-run cap high enough not to kill healthy runs (>22 GB) leaves under 10 GB for everything else and barely constrains anything — so a per-run cgroup cap is a weak instrument on this host, and it is not what would have stopped the Aug 19 events. Holding for a decision on direction rather than patching the numbers.

abl030:
Scrapped at the operator's call. A per-run cgroup cap is the wrong instrument on this host: the suite legitimately peaks at ~21 GB of a 32 GB box, so any cap high enough not to kill healthy runs barely constrains anything — and the 2026-08-19 incident was six separate OOM events with 0.7–3.2 GB victims, i.e. aggregate pressure a per-run cap would not have bound. Not merging; branch deleted. Findings preserved on #1238.

Issue #1149, opened 2026-08-13
feat(dashboard): partial-album discoverability — surface incomplete albums, one-click re-acquire, completeness overrules quality


abl030:
## Operator intent (verbatim, 2026-08-13 session)

"a dashboard bit much like we have now for beets albums with missing tracks. that seems bad and we should have a button to make those 'wanted' and ensure that full rips overrule partials at any quality level … basically it's discoverability for partial albums"

## Motivation

The #1089 Slipknot case: an 8-track partial (bonus disc tagged with the full release's MBID) lived in the library **since March**, undetected, and only surfaced via an unrelated ambiguity in August. Partials are invisible today — nothing measures an installed album's completeness against its release.

Live measurement (2026-08-13, read-only, per-disc `max(tracktotal)` vs item count): **531 of 8,487 albums (~6.3%) are partial** by their own tags; 0 albums lack tracktotal data. Examples: album 351 "Otoño" 27/54, 1160 "Missing EP" 1/5, 1423 "Performing Parades" 11/40.

## The three parts

1. **Discoverability**: a dashboard card in the Disk Coverage style listing partial albums (have/expected counts). The drift-panel pattern from PR #1139 is the template: server-derived rows, per-row action, `source`-style gating, contract-tested.
2. **One-click re-acquire**: a button that makes the album's exact release `wanted` — for an album whose request is `imported`, return it to the search pool; for a library-only album with no request, create one at the album's exact MBID (strict pressing identity — never a sibling). CLI ⇄ API symmetry as usual.
3. **Policy: completeness overrules quality.** A complete incoming copy must be able to replace an installed partial at ANY quality level (complete opus beats partial FLAC). This is a quality-model change and belongs in `full_pipeline_decision_from_evidence` (the ONE decision point) with the album test set updated FIRST; the existing copy's incompleteness must disqualify its rank/floor standing in the upgrade comparison.

## Open questions (deliberately undecided)

- **Ground truth**: tracktotal is tag data and can lie. Is the tag heuristic good enough for surfacing (probably), and should the *decision* re-verify against the MB mirror at import time (probably — validation already matches the release manifest)?
- **Deliberate partials**: some of the 531 may be curated on purpose (e.g., a single kept disc). The surface likely needs an operator dismiss/accept affordance so the list converges — and dismissals need somewhere durable to live.
[...]

Issue #60, opened 2026-04-10
Codec-aware quality comparison and verified-lossless guardrails


abl030:
## Context

Follow-up from #31. The original issue mixed several historical quality-pipeline bugs. Most spectral/backfill/state items are now covered, but the core design problem remains: quality comparisons still use raw bitrate plus a broad `verified_lossless` bypass.

Current behavior is encoded in `lib.quality.import_quality_decision()` and `quality_gate_decision()`:

- `new.verified_lossless` always imports, even if the final target is a very low bitrate lossy transcode.
- Quality gate accepts any verified-lossless low bitrate by forcing it up to the threshold.
- Cross-codec equivalence is represented only in docs/config, not in the comparison model.

This matters now that `verified_lossless_target` can be Opus/AAC/MP3 targets: Opus 128 should not lose to MP3 320 just because its raw bitrate is lower, but a path like verified FLAC -> Opus 90 should not blindly replace a decent existing album without guardrails.

## Scope

Design and implement a codec-aware comparison model for import/quality decisions. Likely shape:

- Keep raw `min_bitrate` as audit/telemetry, but stop using it as the only cross-codec quality metric.
- Add a normalized quality score or comparable tier for `AudioQualityMeasurement` / final formats.
- Account for codec/final format, verified-lossless source status, CBR/VBR, spectral/transcode state, and configured `verified_lossless_target`.
- Use the model in `import_quality_decision()`, `quality_gate_decision()`, and `full_pipeline_decision()`.
- Add explicit guardrails for verified-lossless target formats so very low target bitrates do not bypass all downgrade checks.

## Acceptance criteria

- Opus 128 from verified lossless compares as equivalent/better than MP3 V0 / MP3 320 for the pipeline's purposes.
- A deliberately too-low verified-lossless target, e.g. Opus 90 or equivalent, does not blindly replace a good existing album without policy approval.
- `verified_lossless=True` remains enough to trust genuine lo-fi V0 cases, but not enough to erase all target-format guardrails.
- Existing raw bitrate fields remain available for display/audit.
- Pure decision tests cover MP3 V0, MP3 CBR, Opus 128, lower Opus/AAC targets, and verified-lossless lo-fi cases.
- Simulator scenario tests cover at least one cross-codec replacement case.

## Related

- #31
- #16
- #36
- docs/opus-encoding.md

PR #1251, opened 2026-08-23
feat(quality): an incomplete installed copy must never make a complete candidate a downgrade (#1241)

### LINKED ISSUE #1241: An incomplete installed copy must never make a complete candidate a downgrade
> **Design superseded 2026-08-25.** The settled design below ("stage, never
> auto-replace"; measured completeness; proof-lock grant precondition) is
> replaced by an operator-set incomplete mark driving the normal upgrade
> path — see
> https://github.com/abl030/cratedigger/issues/1241#issuecomment-5403909845.
> PR #1251 is closed; its `scenario_covers_declared_program` conjunct and
> the Dirt Dress pinned world are salvaged into the replacement. The
> verified findings and corrections below remain accurate.

> **Rewritten 2026-08-22 after verification.** The original framing of this issue
> ("teach the whole system about missing tracks at every point… it's a big job")
> did not survive contact with the data. Several of its load-bearing claims were
> wrong, the scale was overstated by three orders of magnitude, and the data
> source it proposed to build on is the wrong one. The verified findings and the
> settled design are below. Original text is in this issue's edit history.

## The invariant

> **If the installed copy is incomplete, a complete candidate is never a downgrade.**

That is the whole behavioural change. One predicate, at one decision site.

## The live incident

2026-08-22 05:27 AWST, request 1852, Dirt Dress — *Theme Songs* (Discogs `4738671`).

The installed copy is four files totalling **719.4 s** against a declared program of
**1418 s**. `04 Iguana  _ Peter and the Wolf.m4a` measures 181.4 s where the two
declared components total 881 s: the 11:40 B-side has been absent since March.

After #1237 cleared the sub-position arithmetic, the pipeline finally fetched a
**complete** rip — five files, 1419.9 s, including `05 peter and the wolf.mp3` at
700.4 s (peer `iosononessuno`, `download_log` 40355).

```
05:27:07  Download complete — 5 files, incl. `05 peter and the wolf.mp3` @ 700.4s
05:28:22  BEETS_VALIDATE  valid=False  scenario=high_distance  distance=0.1667
05:28:23  Curated move → wrong_matches/Dirt Dress - Theme Songs (2007) [b64e0a71]
   …      Wrong Matches triage → action="deleted_reject"  reason="downgrade"
```

Recorded triage reasoning:

```json
"comparison_basis": {"branch": "cross_family_same_rank",
  "existing_format": "aac", "existing_rank": "good", "existing_value_kbps": 128,
  "new_format": "mp3",      "new_rank": "good",      "new_value_kbps": 196,
  "verdict": "equivalent"},
"preview_decision": "downgrade", "preview_verdict": "confident_reject"
```

Both copies are ~128 kbps-grade transcodes off a 16.5 kHz cliff, so on quality they
*are* equivalent. The question never asked: the installed copy is missing 49% of the
album's runtime and the candidate is complete.

The artifact was deleted, the peer was permanently banned for the request
(`source_denylist`, reason `beets validation rejected`, no removal path exists —
the only `DELETE` is a CLI repair restricted to `spectral%` reasons), and the
request returned to `wanted`. It now stands at 350 search attempts.

**The system already knew.** The daily census snapshot generated at
2026-08-21T16:12:47Z — five hours and fifteen minutes before the deletion —
already contained:

```json
{"artist": "Dirt Dress", "title": "Theme Songs", "release_id": "4738671",
 "findings": [{"kind": "missing_source_audio", "detail": "Peter and the Wolf"}],
 "source_audio_components": 5, "physical_audio_files": 4}
```

It wrote down the right answer and then a different code path destroyed the fix.

## What is actually missing

Exactly one thing: **the decision chain has no idea what is on disk is incomplete.**

- `full_pipeline_decision_from_evidence` (`lib/quality/pipeline.py:1503`) takes two
  `AlbumQualityEvidence` rows and `AlbumQualityEvidenceDecisionFacts`. Between them
  that is 53 columns and 8 fields; none is completeness.
- The Wrong Matches cleanup reducer consumes that same verdict, so it inherits the
  blindness.
- The census computes the fact correctly, daily, and writes it to
  `/var/lib/cratedigger/library-completeness.json`, which only
  `web/routes/pipeline_dashboard.py:149` reads.

**The candidate side needs nothing.** `lib/beets.py` already refuses any candidate
missing a declared track:

```python
n_extra = len(candidate.extra_tracks)
if n_extra > 0:
    result.valid = False
    result.scenario = "extra_tracks"
```

So every candidate that reaches the quality comparison is complete by construction.
Completeness is not a two-sided comparison — it is a single predicate on the
installed copy.

## The change

1. **Persist installed completeness** where the decider can read it, sourced from
   `lib/library_completeness.py` — not from `album_tracks` (see corrections below).
2. **One predicate in `full_pipeline_decision_from_evidence`**: an incomplete
   installed copy cannot make a complete candidate a downgrade. The importer and
   the cleanup reducer share this function, so both paths are fixed at once.
   Two conjuncts are required — the installed copy is *positively* incomplete,
   AND this attempt proved its candidate covers the whole declared program.
   It fires on **both** `equivalent` and `worse`: completeness is
   lexicographically above quality, so a copy holding the whole program
   outranks one that does not at every quality level. That does not weaken
   issue #60's "worse is blocked regardless", which governs *import* — the
   hold imports nothing.
3. **Stage, never auto-replace.** A complete candidate against an incomplete
   installed copy lands in `wrong_matches` under a new reason code for operator
   confirmation. It is not imported automatically — a complete-but-lower-quality
   copy crossing codec families is an operator call.
4. **Completeness becomes a precondition on granting the verified-lossless proof
   lock.** The lock's semantics are untouched; an incomplete album simply never
   enters the locked state. Blocked only on a positive incomplete verdict, never
   on `unknown`. Note the site: `lib/quality_evidence.py` ~904 mints the proof
   for the *candidate's* own bytes, but the lock reads
   `current.verified_lossless_proof`, written by
   `propagate_candidate_evidence_to_current` and
   `backfill_current_evidence_from_album_info`. The withhold belongs there, and
   is scoped to a *changed* snapshot fingerprint so a rebuild at the same
   content address never revokes an existing proof.

Per `.claude/rules/code-quality.md` this ships as a pin + generated property pair
with the album test set updated first, and the Dirt Dress world pinned as a
live-incident scenario: 1418 s declared, 719 s installed, complete candidate,
equivalent bitrate, currently `deleted_reject`.

## Explicitly out of scope

The original issue proposed a change at eight decision points. Seven are
unnecessary:

| Proposed | Verdict |
|---|---|
| Never delete a more-complete candidate (cleanup reducer guard) | **Unnecessary** — the decision no longer calls it a downgrade, so there is nothing to guard against |
| Never ban a peer who supplied a more-complete rip | **Reversed — the ban is KEPT, deliberately.** See "The ban stays" below |
| Prefer more complete candidates at enqueue | **Moot** — incomplete candidates are refused outright |
| Search/upgrade intent: a partial is always wanted | **Dropped** — deliberately not changing search intent |
| Candidate completeness measurement in preview | **No new measurement** — but a guard IS required; the "complete by construction" premise is false in the reducer lane, see below |
| Ground-truth / duration-coverage measure | **Dropped** — the existing identity-based functions are better (see below) |
| Dismiss affordance for deliberate partials | **Not needed** — with search intent unchanged, a deliberate partial merely gets a complete copy staged for you to decline |

Also settled: **incomplete candidates stay refused outright.** One incomplete copy
must never "upgrade" another. Cratedigger stays release-based.

### The ban stays

The original rewrite of this issue said the peer must not be banned. That is
**reversed**, deliberately.

The ban in the live incident was not written by the cleanup reducer — the
reducer writes no denylist row at all. It comes from `album_source.py`'s
`reject_and_requeue` at beets-reject time, before any evidence exists, and is
authorized by the beets reject itself
(`lib/world_invariants.py::derive_denylist_authorities`). Suppressing it would
need a second, pre-evidence completeness decision site, breaking "quality
decisions live in ONE place"; it would break
`tests/test_wrong_match_routing_generated.py`'s oracle; and it would remove the
only thing stopping the next cycle re-downloading the identical copy from the
identical peer while the good copy sits in the worklist awaiting confirmation.

It is also harmless here: the denylist is `(request_id, username)`-scoped, and
force-importing the kept row is unaffected by it.

So the required outcome is **kept + banned + visible** — which is #1077's
invariant verbatim (`kept iff its contributing peers are denylisted, kept
implies visible in the worklist`), requiring no extension of that contract.

### The "complete by construction" premise was false

This issue previously asserted that beets' `extra_tracks` reject means every
candidate reaching the quality comparison is complete. That holds in the
importer lane but **not in the cleanup reducer**, because `extra_tracks` is
itself one of the four scenarios admitted to the delete-eligible reducer
(`lib/wrong_match_policy.py::DELETE_ELIGIBLE_REJECTION_SCENARIOS`). The reducer
therefore routinely runs the comparison on candidates beets rejected precisely
*for missing tracks*, and the naive predicate would have staged one incomplete
copy over another — the exact thing this issue forbids.

The fix needs no new measurement: coverage is derived once from the attempt's
already-persisted beets scenario
(`download_log.validation_result->>'scenario'`) via a shared predicate used by
both the reducer and the force lane, and carried as an action-time fact on
`AlbumQualityEvidenceDecisionFacts`.

## Corrections to the original issue

**1. "181 s of an 881 s album — 79% of the runtime is missing."**
881 s is the coalesced *component* (Iguana 181 + Peter and the Wolf 700), not the
album. The album declares 1418 s; the installed copy measures 719.4 s. The correct
figure is **49% missing**, not 79%.

**2. "That loop is the shape behind 12,040 permanent peer bans."**
`source_denylist` holds 18,274 rows, of which 12,051 carry reason
`beets validation rejected`, spread across **3,648 distinct requests**. The three
albums the census flags as incomplete account for **three bans in total** (Dirt
Dress 1, Stellastarr\* 0, UNKLE 2). The top accumulator is request 4082 (Styrofoam,
135 bans) — an ordinary wrong-pressing case. This loop explains ~0.02% of the ban
population. The mechanism is real; the scale claim is not.

**3. "`album_tracks` … the data has been sitting in the pipeline DB since the
request was created, and nothing consumes it."**
The coverage claim is exact — 94,464 rows, 8,564 of 8,564 requests. **But
`album_tracks` is not a usable completeness manifest.** It records disc number,
track number, title, duration and artist. It does not record whether a component
is audio.

Measured across the library: 8,481 of 8,564 requests matched a Beets album, 8,145
measurable (336 excluded for NULL declared durations). Agreement is extremely tight
— median 0.0 s, p99 +0.2 s — and 19 albums show a shortfall of ≥30 s or ≥2%. Of
those 19, **one is a genuine hole and 17 are false positives**:

- **8** — a whole declared disc absent, every one a bonus DVD. Prince *Sign o' the
  Times* disc 9 (26 tracks / 8,651 s), Sufjan *The BQE* disc 2 (one 3,009 s film
  track), Rufus Wainwright, Efterklang, Emery, Miles Davis, Anberlin, Libertines.
  In seven of the eight the residual is exactly zero.
- **3** — video inline on an audio medium. Beatles *Now and Then* (a 34 s short-film
  trailer), TV on the Radio *Chemical Peels* (a 203 s music video), Skrillex *More
  Monsters and Sprites* (MusicBrainz `recording.video = true` on track 8).
- **1** — Bouncing Souls *Anchors Aweigh*: Discogs overlapping-duration convention,
  the hidden track is inside the 577 s file.
- **1** — tUnE-yArDs *w h o k i l l*: the manifest lists `Bizness` twice.
- **1** — …soihadto…: the file is present on disk and merely uncatalogued in Beets,
  behind a stale zero-byte `.beets` temp file. The census reports this correctly as
  `catalog_drift`.
- **3** — declared-duration slop (Alexi Murdoch 1.4%, Sleep Token 2.5%, Kerrie
  Biddell 2.3%), all with matching track counts.

Real: Dirt Dress. Ambiguous: Stellastarr\* (a hidden track behind 20 s of declared
silence). **A ~95% false-positive rate**, and under this issue's invariant those
false positives would disqualify correct albums' rank and make them permanently
unsatisfiable.

**4. The census is not a lesser instrument — it is the correct one.**
`lib/library_completeness.py` already excludes non-audio components via
`recording.video`, reproduces Beets' own Discogs sub-position coalescing
(`discogs_manifest`), reconciles the Beets catalog against the filesystem, and
returns a genuine `unknown` rather than guessing. Its four current findings are
right, with UNKLE the one unproven case. It needs no measurement work — only
somewhere to put its answer.

**5. #1237's `composite_audio_gap.py` must not become a decision input.**
It is documented "Never a decision by itself" and has a false-alarm direction: a
segued or medley recording with every part present reads as `missing_source_audio`.
Two of the census's three `missing_source_audio` findings are that shape.

## Resolved open questions

| Question | Resolution |
|---|---|
| Is duration-coverage the right primary measure? | No. Use the census's identity-based comparison. Duration arithmetic measured at ~95% false positives. |
| Threshold — any missing component, or a floor? | Dissolved. The census returns a categorical verdict; there is no number to threshold. |
| Does a complete copy overrule a proof-locked partial? | Reframed: completeness becomes a precondition on *granting* the lock. Lock semantics unchanged. |
| Replace vs stage? | Stage, under a new reason code. |
| Deliberate partials | No affordance needed; search intent is unchanged. |

**Retroactive lock revocation: no action.** 7,229 of 8,564 requests (84.4%) are
proof-locked, and UNKLE is the only one currently flagged incomplete. Its manifest
declares `Nursery Rhyme` at 285 s and `Breather` with no duration at all — the
Discogs sub-position convention — against a 265.5 s physical file. Both readings
leave the same ~19.5 s discrepancy, so the metadata cannot discriminate, and the
"no silence gap" verdict is exactly the medley false alarm `composite_audio_gap.py`
documents. Not proven incomplete; the lock stays.

## Follow-ups (not blockers)

- Dirt Dress request 1852 still has its only known complete source permanently
  banned. Clear that one `source_denylist` row **after** this ships, not before.
- …soihadto… has a stale zero-byte `.beets` temp file and one uncatalogued mp3;
  a Beets re-scan fixes it with no re-download.
- Stellastarr\* and UNKLE both need thirty seconds of listening to settle.

## As built

Branch `feat/1241-installed-completeness` (`0d8fb550`, `e3649ffd`).

- **Migration 082** adds `album_quality_evidence.installed_completeness` (JSONB,
  shape-checked). NULL means *never measured* and reads as `unknown`; a stored
  `unknown` means measurement was attempted and could not decide, and is kept
  distinguishable so a dead mirror is not re-fetched every cycle. Both behave
  identically for policy. Forward-only, no backfill — rows fill on their next
  preview cycle.
- Preview measures it with `lib/library_completeness.py`'s classifier, the same
  one the #1149 census runs, with the boolean composite-silence-gap instrument
  deliberately not consulted.
- `DECISION_INSTALLED_INCOMPLETE_HOLD` is deliberately **not** a member of
  `QUALITY_DECISION_REJECT_STAGE_DECISIONS`; that omission is what makes the
  reducer keep the folder.

**Named residuals**, written up in `docs/quality-verification.md` rather than
left implicit:

- The "folder is kept" outcome is the **cleanup reducer lane only**. The
  auto-import lane still deletes its disposable staged source and bans, exactly
  as before. Widening that needs a new quality-reject-writes-`failed_path`
  quarantine lane, which `docs/rejection-routing.md` currently rules out.
- The write-time proof-lock precondition cannot make "no locked album is
  incomplete" a standing invariant. Three cohorts remain: granted before the new
  snapshot has any verdict, carried from a complete predecessor, and pre-deploy
  rows. Closing it fully requires the lock's *read* side to become
  completeness-conditional.
- The reducer's verified-lossless-parent short circuit still returns before the
  quality decision, so it deletes without consulting completeness — the reducer
  half of the same cohort.

**Evidence.** Eleven planted mutants confirmed RED in both directions, including
re-narrowing the hold to `equivalent` only (3 RED: both hand-written twins and
the generated parity property). Full gate green across all six phases.

## Related

- #1149 — census (part 1) shipped and is the correct source; this supersedes part 3.
- #1237 — sub-position arithmetic and the composite instrument (shipped 2026-08-21).
- #1183 — the original indexed-subtrack audio loss.

abl030:
Closes #1241.

## The invariant

> **If the installed copy is incomplete, a complete candidate is never a downgrade.**

Request 1852 (Dirt Dress — *Theme Songs*) lost a complete rip to exactly this:
719 s installed against a 1418 s declared program, a complete 1419.9 s
candidate, equal bitrate grade → `verdict: "equivalent"` → `deleted_reject`,
and the only peer holding the missing 11:40 B-side permanently banned for the
request. The daily census had already written down `missing_source_audio:
"Peter and the Wolf"` five hours and fifteen minutes earlier; nothing in the
decision chain could read it.

## What changed

- **Migration 082** — `album_quality_evidence.installed_completeness` (JSONB,
  shape-checked). NULL means never measured and reads as `unknown`; a stored
  `unknown` means measurement was attempted and could not decide, kept
  distinguishable so a dead mirror is not re-fetched every cycle. Both behave
  identically for policy. Forward-only, no backfill.
- **Measurement** — preview runs `lib/library_completeness.py`'s classifier,
  the same one the #1149 census uses, with the boolean composite-silence-gap
  instrument deliberately not consulted (it is documented "never a decision by
  itself" and has a medley false-alarm direction).
- **One predicate** in `full_pipeline_decision_from_evidence`, shared by the
  importer and the Wrong Matches cleanup reducer.
  `DECISION_INSTALLED_INCOMPLETE_HOLD` is deliberately not a member of
  `QUALITY_DECISION_REJECT_STAGE_DECISIONS` — that omission is what makes the
  reducer keep the folder for the operator instead of deleting it.
- **Proof lock** — completeness becomes a precondition on *granting* the
  verified-lossless lock. Lock semantics untouched; an incomplete album simply
  never enters the locked state. Scoped to a changed snapshot fingerprint so a
  rebuild at the same content address never revokes an existing proof.

The hold covers **both `equivalent` and `worse`**. Completeness is
lexicographically above quality — a copy holding the whole program outranks one
that does not, at every quality level. This does not weaken issue #60's "worse
is blocked regardless", which governs *import*: the hold imports nothing, it
withholds a destructive reject and stages the folder. A 96 kbps complete rip
against an incomplete 320 kbps album becomes a worklist entry a human confirms,
never a silent replacement. The comparison basis keeps its honest `verdict`, so
a held world still audits as `worse`.

## Two corrections found during implementation

**The "complete by construction" premise was false.** `extra_tracks` is itself
one of the four scenarios admitted to the delete-eligible reducer, so the
reducer routinely runs the quality comparison on candidates beets rejected
*precisely for missing tracks*. The naive predicate would have staged one
incomplete copy over another. Fixed with a coverage conjunct derived once from
the attempt's already-persisted beets scenario via a shared predicate used by
both the reducer and the force lane — no new measurement.

**The peer ban is kept.** The ban is written at beets-reject time, before
evidence exists, and is authorized by the reject itself. Suppressing it would
need a second pre-evidence decision site, break the routing oracle, and remove
the only thing stopping the next cycle re-downloading the identical copy from
the identical peer while the good copy waits in the worklist. Outcome is
**kept + banned + visible**, which is #1077's invariant verbatim.

## Named residuals

Documented in `docs/quality-verification.md` rather than left implicit: the
"folder is kept" outcome is the cleanup-reducer lane only (auto-import still
deletes its disposable staged source); the write-time lock precondition cannot
make "no locked album is incomplete" a standing invariant (three cohorts
remain); and the reducer's verified-lossless-parent short circuit still returns
before the quality decision.

## Verification

Album test set updated first, with the Dirt Dress world pinned as a
live-incident scenario on both twins. Pin + generated property pair per
`.claude/rules/code-quality.md`.

**Eleven planted mutants, all confirmed RED in both directions** — including
re-narrowing the hold to `equivalent` only (3 RED: both hand-written twins and
the generated parity property), dropping either completeness conjunct, and
disabling both measurement call sites.

Full gate green across all six phases (js-syntax, js-unit, pyright, ruff,
vulture, python), run independently of the implementing agents.

## Deployment note

Migration 082 is not yet deployed. Every existing row reads NULL → `unknown` →
no behaviour change, so the change is inert until the column is measured.
Verified against the live DB: zero rows currently carry the new derived output.

After this ships, request 1852's remaining `source_denylist` row can be cleared
so the complete copy can be re-fetched — **not before**, or the next cycle
simply repeats the deletion.


abl030:
Closing as superseded by the operator's 2026-08-25 design reversal — https://github.com/abl030/cratedigger/issues/1241#issuecomment-5403909845. The measured-completeness column (migration 082), preview-time census measurement, `DECISION_INSTALLED_INCOMPLETE_HOLD` staging lane, and proof-lock grant precondition are all dropped in favor of an operator-set incomplete mark on `album_requests` driving the normal upgrade path. Salvaged from this PR into the replacement: `scenario_covers_declared_program` + the `_attempt_beets_scenario` threading, and the Dirt Dress pinned world.

PR #1257, opened 2026-08-25
feat(#1241): operator incomplete mark — complete always beats incomplete

abl030:
## Summary

Issue #1241 (non-closing until live proof), replacing the closed PR #1251's
measure+stage+proof-lock design per the operator's reversal
(https://github.com/abl030/cratedigger/issues/1241#issuecomment-5403909845 —
"incomplete is incomplete and complete always always beats it").

**The mark.** Migration 082 adds `album_requests.marked_incomplete_at` —
operator-owned, never written by measurement; the daily library-completeness
census only informs the decision. One canonical service
(`lib/incomplete_mark_service.py`) behind both surfaces:
`pipeline-cli mark-incomplete <id> [--clear]` and
`POST /api/pipeline/mark-incomplete`, plus a Mark incomplete / Clear button
on the dashboard's Library Completeness card (server-side enrichment joins
each embedded census album to its request + mark state).

**The predicate.** `installed_marked_incomplete AND
candidate_covers_declared_program` — when both hold,
`full_pipeline_decision` (and its evidence twin) disregards the installed
side entirely: every `existing_*` input, the one-sided spectral floor, the
lossless-source anchor, and the decision-21 verified-lossless lock. The
candidate is admitted exactly as into an empty slot; the generated parity
property pins this as an exact dict equivalence against the same world with
no current album (modulo the audit-only `installed_incomplete_disregarded`
key), and pins byte-identity to the unmarked baseline for every world where
the predicate does not hold. Monotone by construction. Absolute
candidate-side floors still apply, and a below-par import keeps the search
open, so quality convergence resumes against the now-complete copy.

**The coverage conjunct** (salvaged from PR #1251) is derived in ONE place —
`lib/validation_envelope.py::scenario_covers_declared_program` — from the
attempt's persisted beets scenario: `strong_match`/`high_distance`/
`unmapped_audio` are assigned only after `apply_candidate_scenario`'s
`extra_tracks` branch fell through (verified against `lib/beets.py`, not
taken from the brief). `extra_tracks` proves the opposite: one partial can
never "upgrade" another. Auto/local import is covered by admission; a force
import reads the linked `download_log` row's stored scenario
(`_attempt_beets_scenario`) — the same bit the Wrong Matches cleanup
reducer derives from that row, so the two lanes cannot disagree.

**Reducer lane.** A marked request's covered Wrong Matches row now decides
import-class, so the reducer KEEPS the folder (kept + banned + visible per
#1077) instead of deleting it as a "downgrade" — the Dirt Dress incident's
exact deletion path — and the verified-lossless-parent short circuit yields
for the same rows.

**Auto-clear.** A terminal acceptance whose candidate was proven whole
clears the mark through the imported transition's metadata CAS
(`_do_mark_done` → `to_imported_fields`), atomic with the status write.
Rejections and preserve-imported outcomes leave it untouched.
`to_imported_fields` silently drops unthreaded keys, which the
orchestration test caught RED before the plumbing landed — the exact
silent-drop shape Rule A exists for.

**Tests.** Album test set first (Dirt Dress request 1852 / download_log
40355 pinned through both twins, with lock-disarm, worse-but-complete, and
no-regression corners), the generated parity property above, reducer
keep/delete/shortcut pins, force-lane and automation-lane orchestration
(flip + clear + preserve), Rule A real-PG round-trip, fake self-tests, CLI
exit-code and API status-code contracts, dashboard enrichment contract.
Typing-ratchet convergence: `tests/helpers.py` gains the
`dispatch_import_with_fake_db` bridge and `test_dispatch_core.py`'s
`type_ignore` count goes 10 → 0 (net −10 escape hatches; +3 bridge `Any`
per the `finalize_claimed_dispatch` precedent).

**Rule D / live-corpus evidence — the vacuous case.** The affected
population is `album_requests WHERE marked_incomplete_at IS NOT NULL`,
which is zero rows by construction: the column is created by this PR's own
migration, so no pre-deploy row can carry it, and no unmarked row's
decision can move — the property's negative half pins byte-identical
decision dicts for every world without the predicate, and the decision
dict's one new key is audit-only (no branch reads it; `decision_differential`'s
derived DECISION_KEYS picks it up automatically). Per
`.claude/rules/test-fidelity.md` § "The vacuous case", the population
argument stands in for a differential that would be zero by construction.

**Visual evidence.** Dev-server screenshot loop (local migrated ephemeral
PG + synthetic census snapshot, since the live DB lacks migration 082 until
deploy): the Library Completeness card renders all three states — a marked
request's row shows "Clear incomplete mark", an unmarked one "Mark
incomplete", and a census album with no resolvable request shows no button;
clicking against the GET-only dev server fails closed with an honest toast
and the button restored. Live click-through is part of post-deploy
verification.

## Fault injection

Nine planted mutants, all KILLED (driver run with
`PYTHONDONTWRITEBYTECODE=1`, clean-tree killer suite green after):
dropping the coverage conjunct from the flat-twin predicate (killed by the
uncovered-candidate pin AND the parity property's corner `@example`);
dropping the evidence twin's lock-guard disarm (evidence lock pin);
keeping `current_verified_lossless_proof` or `existing_min_bitrate` under
the disregard (flat lock pin / Dirt Dress fresh-equivalence pin); dropping
the reducer facts threading (marked-kept reducer pin); dropping the
dispatch mark threading (force-lane flip test); dropping the `_do_mark_done`
clear and, separately, the `to_imported_fields` field threading (automation
clear test); dropping the reducer VL-parent shortcut disarm (killed by the
new shortcut pin — an earlier campaign draft showed this mutant SURVIVING,
which is why that pin exists).

## Risk

- The decider changes NOTHING for unmarked requests — machine-pinned as
  byte-identical decision dicts by the property's negative half, and the
  live marked population is zero until an operator marks something.
- The mark authorizes automatic replacement of the installed copy by any
  admissible beets-whole candidate at any quality, including over a
  verified-lossless proof — that is the operator's explicit intent (the
  mark IS the authorization), and the post-import gate keeps searching
  after a below-par replacement.
- Deploy-window: migration 082 is additive; old code reading the new
  column is safe (`msgspec.convert` ignores extra keys). New code against
  the un-migrated schema would fail `AlbumRequestRow` validation, but unit
  ordering runs the migrate oneshot before the workers start.
- Not changed: search intent/cadence, denylist policy, #1077's
  kept⟺banned invariant, the beets identity validation (a complete
  candidate that fails distance still needs a one-click force import),
  proof-lock GRANT semantics, and every existing wrong-match routing row.


abl030:
## Independent fable-tier review + correction round (eda0f2bc)

A fresh-context fable reviewer ran the full obligation set (claim re-derivation, namespace cross-reference, 22 planted mutants aimed at the new tests, docs-by-artifact grep, archivist-frame audit). Verified clean: the scenario-ordering claim in lib/beets.py, the harness action-file lane never re-deciding, all 11 existing-side inputs nulled under the disregard, the metadata-CAS clear atomic on both terminal paths (rescue included), the decision-dict and column namespaces, and the Rule-D vacuous-population argument.

Findings, all fixed in eda0f2bc:

- **F1 (bug, would have shipped):** `preflight_existing` — the keep-the-existing-import outcome with nothing installed — cleared the operator's mark. Now carved out exactly like `clear_stale_v0_probe` on the same call, with a pin.
- **F2/F3/F5/F6 (gaps proven by surviving mutants):** preserve-on-uncovered-acceptance pinned; the JS button got real three-state + exact-handler-args assertions (an argument-inversion mutant had survived all 414 existing JS assertions); the `lossless_source_locked` disarm got a deterministic pin + promoted parity `@example` (previously fuzz-tier-only); the unmarked force twin now fails at the quality decision (code asserted), not launch authority.
- **F4:** false sentence in `docs/rejection-routing.md` corrected (`unmapped_audio` is not delete-eligible).
- **F7:** `updated_at` restamp asserted through real PG and mirrored in the fake; the early-reject audit-flag claim pinned; the CLI live-replay's marked second decision pinned via a #409 narrow protocol (no new typing escape hatches).
- **F8:** preview-lane asymmetry documented as deliberate rather than half-threaded.
- **F9:** dashboard enrichment gained per-album error isolation (one refused presentation row no longer 500s the route), tested with the real refused shape.

Round-2 fault injection: all five previously-surviving mutants now KILLED (preflight carve-out drop, unconditional clear, existing-probe keep, dispatch marked-hardcode, JS inversion); clean tree green. Fresh final-gate receipt: pass at eda0f2bc.

PR #1262, opened 2026-08-25
fix(#1261): Discogs sub-position, heading-row, and vinyl-grammar tracklists become rip-shaped manifests

### LINKED ISSUE #1261: Discogs sub-position tracklists produce unmatchable manifests — 28 wanted requests can never pass the matcher count gate
## Symptom

Surfaced from the #1149 census deferral: requests 6188 (Stellastarr* — Harmonies For The Haunted) and 3814 (Marie Wilson — Heartbreak) have hundreds of search attempts and zero download attempts. Investigation (2026-08-25) proved the class is wider: **28 currently-wanted Discogs requests carry sub-position track debris**, including Kid A (358 attempts), Icky Thump, Aerial, and Thriller Special Edition — albums that are trivially findable, so scarcity is ruled out.

## Root cause

Discogs encodes hidden tracks as sub-positions (`10.1 Island Lost At Sea / 10.2 (silence) / 10.3 Untitled`). `_parse_position` (`web/discogs.py`) has no case for `N.M` positions — they fall through to `(disc 1, track 0)` — and the tracklist loop in `get_release` keeps every sub-entry as its own manifest row, including a literal `(silence)` "track". The persisted `album_tracks` manifest for release 521474 therefore has **12 expected tracks** while every real-world rip of the CD has **10 files** (hidden track baked into the last file).

The matcher (`lib/matching.py`) requires `file_count == total_tracks` AND `matched_tracks == total_tracks`, so every real copy is rejected forever:

- `strict_count_mismatch`: 10-file folders vs 12 expected — includes a 16/44.1 FLAC copy rejected every cycle (search_log 485322, peer `Critical Damage`).
- `avg_ratio_low` at 11/12: one peer's rip splits the hidden track (12 files); the sole `missing_title` is `(silence)` — a 20-second gap no filename can ratio-match.
- Cohort signatures match: Kid A best candidate 12/14, Thriller SE 21/22, Icky Thump 17/17 filenames matched then title cross-check failed.

Live slskd searches confirmed 8+ complete Stellastarr* copies (MP3 320/V0 + two FLAC rips), all 10 files. The pathology blocks at the **matcher**, upstream of validation — the census's "not the sub-position pathology" call only covered the validation side.

A second flavor in the same parser: bare vinyl side positions (`A`, `B` with no digit) also fall through to track 0, scrambling manifest ordering for singles (e.g. requests 1852, 8840 with every track at 0).

## Fix (this issue)

Fix the data, not the gate — the strict count gate is load-bearing pressing-identity enforcement.

1. **Ingestion normalization** in `web/discogs.py::get_release`'s track loop:
   - Group `N.M` sub-positions into a single track N: title from the first non-silence sub-entry, durations summed. Mirrors how MusicBrainz models the same discs ("Island Lost at Sea / [untitled]", 7:57 — matching the real 19 MB MP3 / 56 MB FLAC files).
   - Bare letter positions (`A`, `B`) parse as (side-as-disc, track 1), consistent with the existing `A1`/`B1` convention.
   - `get_release_raw` (source-audit consumers) keeps literal positions untouched.
2. **Tests first**: deterministic pins for the 521474 and 18213658 shapes, bare-letter singles, and unchanged flat/`A1`/`1-3` forms, plus a generated property (one manifest entry per top-level position group, durations preserved, flat lists unchanged) per the pin+property pair rule.
3. **One-shot backfill** of the 28 wanted Discogs requests during the deploy window (agent-run per scope.md, never committed): re-fetch each release through the fixed normalizer, rewrite `album_tracks`, supersede the active search plan so the next cycle regenerates queries from the corrected manifest. No `SEARCH_PLAN_GENERATOR_ID` bump — the generator algorithm is unchanged; only its DB inputs move, and only for the backfilled requests.

## Explicitly out of scope

- Loosening the matcher count gate or title cross-check (would admit sibling pressings).
- Request 3814 acquisition: Marie Wilson's catalogue is genuinely absent from Soulseek today (album query: 0 results; her only network presence is two loose 1998 compilation tracks). The manifest fix makes the perpetual search *able* to succeed; YouTube rescue remains the realistic path meanwhile.
- Request 6188 needs no special handling once this lands — copies are abundant.

## Affected wanted cohort (measured 2026-08-25)

113 requests have a `(silence)` manifest track overall; 30 active / 28 wanted+Discogs carry track-0 rows: 1852, 1858, 1859, 1868, 1897, 2370, 2684, 3133, 3814, 4223, 4946, 5028, 5096, 5115, 5282, 5400, 5401, 5402, 5438, 5936, 6188, 6937, 8836, 8838, 8840, 8849, 8851, 8982.

abl030:
Fixes #1261.

## What

Discogs tracklists reach `album_tracks` in shapes no rip can match, and the matcher's strict count gate (`file_count == total_tracks`) then rejects every real copy forever. Three flavors, all measured live on the 28-request wanted cohort:

1. **Sub-position hidden-track runs** (`10.1 Song / 10.2 (silence) / 10.3 Untitled`): every sub-entry became its own manifest row (at track 0), so a 10-file rip faced a 12-track manifest. Requests 6188 (Stellastarr*), 3814 (Marie Wilson), 5096 (UNKLE).
2. **Empty-position heading rows** (Discogs side/disc/bonus section labels the mirror flattens into the tracklist): Kid A's vinyl carried 4 side-name headings → 14 expected tracks for a 10-file rip. This is the cohort's dominant flavor (Kid A, Icky Thump 17→13, Aerial 18→16, Thriller SE 22→21, Eluvium's box set 47→41).
3. **Unparsed vinyl grammars** (bare `A`/`B`, `1A/2A/1B`, trailing-dot `1.`): rows landed at the `(1, 0)` sentinel, scrambling manifest ordering and the resolver's positional `track_artist` mapping.

`lib/discogs_positions.py` is now the one canonical parser + normalizer, consumed by **both** production paths — `web/discogs.py::get_release` (add/Replace/CLI persist callers and browse display) and `album_source.py::_populate_tracks_discogs` (the search worker's empty-manifest fallback, which previously carried its own divergent inline parser with no sub-position or vinyl-letter handling at all). Normalization: one row per physical position, first-appearance order; sub-groups keyed on the literal base string (unparseable bases stay distinct); merged title = first non-placeholder sub-entry (blank counts as placeholder); known durations summed; a flat index parent joins its group with authoritative title and total duration; heading rows (empty position AND empty duration, per `lib/library_completeness.py`'s measured mirror rule) dropped only when the release positions its other tracks. The release cache key bumped v2→v3 so warm pre-fix manifests age out; `pipeline-cli add` now passes `fresh=True` like the web add path.

## Evidence (issue #1261 has the full investigation)

- Request 6188: live slskd searches found 8+ complete copies including two FLAC rips; `search_log` 485322 shows a 16/44.1 FLAC candidate rejected `strict_count_mismatch` (10 files vs 12 expected) and an 11/12 near-miss whose only `missing_title` was `(silence)`.
- Kid A: 358 search attempts, zero downloads, best candidate 12/14 matched; its Discogs positions are `['', 'A1', 'A2', '', 'B1', ...]` — 4 headings.
- MB models the same Stellastarr* disc as 10 tracks ("Island Lost at Sea / [untitled]", 7:57) — the normalized manifest now matches that shape.

## Rule D — live-corpus tally

The affected population is the 28-request wanted Discogs cohort (list in #1261) plus any future Discogs fetch. Rendering every cohort release's live mirror tracklist through the old shape (one row per raw entry) vs the fixed normalizer: **26/28 releases change manifest shape** — row-count corrections on 15 (3 sub-position merges: 6188, 3814, 5096; 12 heading drops), numbering-only corrections on 11 (9 bare-letter singles, plus 1852's `1A` grammar and 8840's trailing-dot grammar). Headline rows: Kid A 14→10, Icky Thump 17→13, Aerial 18→16, Thriller SE 22→21, Eluvium 47→41, Stellastarr* 12→10, Marie Wilson 12→10. The two unchanged: 5400 (zero-indexed positions, count already rip-shaped) and 5936, whose tracklist carries a literal `"Video"` position row — an enhanced-CD video track that stays a phantom for audio-count matching (known residual, single request; noted in #1261). Existing persisted rows are untouched by the code (vacuous for the render-differential machinery); the deploy-window one-shot below corrects them, and the browse release-detail view renders the same normalized rows through an unchanged renderer.

## Tests

- **Pins** (`TestGetReleaseSubPositionTracks`, 15 scenarios): the live 521474 shape, Kid A heading shape, leading/all-silence groups, blank-title placeholder, "Silence Kit" anchoring, duration sum/None, parent-total and parent-after-subs authority, empty-position-with-duration kept, all-unpositioned preserved, `sub_tracks` index parent kept, bare-letter/number-letter/trailing-dot grammars, duplicate flat positions kept, flat pass-through.
- **Composition pins** (`TestPopulateTracksDiscogsNormalization`): the REAL fallback writer → real normalizer → `FakePipelineDB.set_tracks` → `get_tracks`, via a new typed bridge (`make_database_source_with_fake_db` in `tests/helpers.py`). The mutant runner proved these two pins are the sole guard on the second writer.
- **Generated properties** (`tests/test_discogs_api_generated.py`): a draw-time STRUCTURAL oracle (expected rows recorded from form definitions at draw, never by re-running production parsing) asserting FULL rows — numbers, titles, durations — over worlds with sub runs, parents, headings, duplicate flat positions, unparseable bases, and blank titles. Strategy limits (placeholder anchoring, parent adjacency) are declared in the module docstring with the deterministic pin that owns each axis.
- RED→GREEN throughout; the original pins failed on the pre-change code.

## Fault injection

Two independent two-agent review rounds (reader + mutant runner, disjoint jobs, runner in its own worktree with `PYTHONDONTWRITEBYTECODE=1`). Round 1: 18 mutants, 17 killed; the survivor (flat rows routed through the group accumulator) became a pin + wider property. Round 2: 18 mutants, 18 killed on the tested surface — but the runner proved the v3 cache-key bump had an ANTI-pin (reverting the bump turned the tree green, R2-M15) and both agents independently caught the stale `v2` literal in `tests/test_web_dev_server.py`; it now pins v3. Round 3: 16 mutants, 16 killed (R2-M15 re-verified killed by the new v3 pin), and the runner proved `test_all_empty_positions_are_preserved`/`_ALL_UNPOSITIONED_PIN` were bystanders for their named `any_positioned` clause (their rows' durations short-circuited the heading predicate) — round 5 de-bystanders both worlds and re-proves the clause mutant RED at the pin (3/3). Round 4's two semantic sites are mutant-qualified directly: coerced-instead-of-raw heading comparison dies at `test_absent_duration_key_heading_shape_is_kept`; reverting the parent-duration truthiness guard dies at `test_zero_duration_parent_does_not_zero_the_group`. Reader findings driving production changes: the parallel `album_source` parser (F2), base-string group keying (F3), flat-parent join (F4), blank-title placeholder (F8), cache-key consumers (F5/D0), parent duration double-count (D2), parent-order title dependence (D3), raw-vs-coerced heading keys (R1), zero-duration parent (R3).

Clause-coverage map (runner's round-3 note, stated rather than implied): first-appearance ordering and the whitespace-placeholder half are property-only (P1); the `sub_tracks` index-parent carve-out and the placeholder anchoring are deterministic-pin-only (both declared in the generated module's docstring as fail-closed legislation / known limits); everything else is pin+property paired.

## Notes for the record

- **Typing ratchet**: `tests/helpers.py` `any` baseline moved 28→30 for the new bridge's own annotations, following the #1246 `dispatch_import_with_fake_db` precedent (one bridge, zero per-call-site hatches). The round-2 reader flags that `.claude/rules/code-quality.md`'s convergence rule can be read as owing a −10 sweep on that file when the ratchet moves; that work is not taken here — operator's call.
- **Bare-letter side-as-disc** (`A`/`B` → discs 1/2): consistent with the existing `A1`/`B1` convention, but it means 7" singles now produce two one-track media and can enter `try_multi_enqueue` like LPs do. Accepted deliberately; noted per the round-2 reader.
- **`1A`-as-side-suffix** diverges from Beets' subtrack-index reading; documented in the module docstring with the count-preservation rationale.
- **(1, 0) sentinel residual**: positions no grammar covers still share `(disc 1, track 0)`; counts stay correct but relative order under `get_tracks`'s `(disc_number, track_number)` ordering is undefined. Recorded in the normalizer docstring.

## Deploy-window one-shot (not committed, per scope.md)

For each of the 28 request ids in #1261: re-fetch the release through the deployed normalizer and rewrite the manifest, carrying existing `track_artist` values over by title match, then regenerate the plan:

1. Heredoc python on doc2: `web.discogs.get_release(discogs_id, fresh=True)` → merge old `track_artist` by title → `PipelineDB.set_tracks(request_id, tracks)`.
2. `pipeline-cli search-plan regenerate <id>` (supersedes the active plan; next cycle searches from the corrected manifest).

Never bulk-delete `album_tracks` rows and let the pipeline re-derive: before this fix that path repopulated through the divergent fallback parser; after it the one-shot is still the correct instrument because it preserves `track_artist`.


abl030:
## Independent review record (two-agent split, three rounds)

**Round 1** (vs 0d0bcc59) — Reader found 11 issues incl. the parallel `album_source.py` Discogs parser (F2, HIGH: the search worker's empty-manifest fallback bypassed the fix entirely and booby-trapped the backfill), base-key merge collisions (F3/F4), the empty-title trap (F8), and the stale-cache window (F5). Runner: 18 mutants, 17 killed; the survivor (flat rows routed through the group accumulator, silently merging distinct flat tracks) drove a new pin + wider property.

**Round 2** (vs d8a498a7) — Both agents independently caught the same blocker: the `v2` cache-key literal in `tests/test_web_dev_server.py` was an **anti-pin** — the runner's R2-M15 (reverting the v3 bump) turned the whole tree green. Runner: 18/18 killed on the tested surface. Reader delta: parent-duration double-count (D2), parent-order title dependence (D3), heading-predicate divergence from `lib/library_completeness.py`'s measured rule, and the `1A`-grammar Beets-divergence disclosure — all addressed in round 3, which also added the cohort's dominant flavor (empty-position heading rows: Kid A 14→10, Icky Thump 17→13, Aerial 18→16, Eluvium 47→41) after live-mirror measurement of all 28 stuck releases.

**Round 3** (vs 215a2f35) — Reader final: no blocker; four LOW findings (raw-vs-coerced heading keys R1, 0:00 parent R3, an undeclared carve-out clause R4, commit-message attribution nits) — all fixed in round 4 with direct mutant qualification. Runner: 16/16 killed, R2-M15 re-verified closed, one finding: the all-unpositioned pins were bystanders for their named `any_positioned` clause — fixed in round 5 with the clause mutant re-proven RED at the pin (3/3).

Final state: canonical suite receipt-backed PASS on 3e2a69b4; every reader finding fixed or explicitly recorded in the PR body (the typing-ratchet convergence question is flagged there for the operator).

